PR #4, opened 2025-09-04
refactor: improve interfaces

SahilVasava:
- created function stamper wrapper and types on top of TurnkeyStamper classes
- added functional client util for kms proxy requests and removed the @turnkey/http package as dep
- added wallet and auth actions for the doorway client


packages/core/src/client/DoorwayClient.ts
```diff
@@ -1,22 +1,27 @@
...
+import type { Stamper } from "../stampers/types.js";
 
 type TurnkeyClientConfig = ConstructorParameters<typeof TurnkeyClient>[0];
 
 interface DoorwayClientConfig extends TurnkeyClientConfig {
   proxyBaseUrl: string;
 }
 
-export class DoorwayClient extends TurnkeyClient {
+export class DoorwayClient<TStamper extends Stamper = Stamper> extends TurnkeyClient {
```
SahilVasava: Will be removing the dependence on `TurnkeyClient`.
We will have our own `client` interface

PR #66, opened 2026-03-26
feat: setup `react-kit` package and demo app

OmarBasem:
### Summary

This PR creates and sets up the following 2 packages:

- `react-kit`: The web UI package. It included 2 example components to show the structure: a `Button` and `CodeInput` components. Each have a test file and a storybook file. The directories can be organized by feature (`auth`, `signing`, etc), and the shared components can be under the `shared` directory.

- `playground`: a playground react app to import and test some of the components and features from `react-kit` and show examples to the developer


packages/react-kit/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# @zerodev/wallet-react-kit
+
+## 0.0.1
+
+### Patch Changes
+
+- Initial release
```
SahilVasava: This file is autogenerated mostly. We use `changesets` to manage publishing. Both core packages are released in alpha mode. So it's not required to create this now. But later on when we are ready to publish `@zerodev/wallet-react-kit@alpha` it will be autogenerated.

packages/react-kit/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+export type { CodeInputProps } from './auth/components/CodeInput'
+// Auth components
+export { CodeInput } from './auth/components/CodeInput'
+export type { ButtonProps } from './shared/components/Button'
+// Shared components
+export { Button } from './shared/components/Button'
+
+export const version = '0.0.1'
```
SahilVasava: This will drift with version bumps. And we will be having alpha versioning before we do regular release. So not sure if it's needed here.

packages/playground/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@zerodev/playground",
```
andrascodes: Wdyt about putting the `playground` app under `apps/playground` instead? It's not a package strictly speaking so it makes more sense for me there.

brtkx: Same, I think all the example/demo apps should be there. Also should we call it something like `apps/react-example`? Maybe we will make a playground in the future and then we can rename.

packages/react-kit/src/shared/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+})
+
+describe('Button', () => {
+  describe('rendering', () => {
+    it('renders with text', () => {
+      render(<Button text="Connect Wallet" />)
+      expect(screen.getByText('Connect Wallet')).toBeDefined()
+    })
+
+    it('renders as a <button> element', () => {
```
brtkx: I don't think this test does anything? `getByRole('button')` in every tests already tests for that inherently.

packages/react-kit/src/shared/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    it('allows overriding type to "submit"', () => {
+      render(<Button text="Submit" type="submit" />)
+      const button = screen.getByRole('button')
+      expect(button.getAttribute('type')).toBe('submit')
+    })
+
+    it('renders without text when text is not provided', () => {
+      render(<Button data-testid="empty-btn" />)
+      const button = screen.getByTestId('empty-btn')
```
brtkx: Consistency question: why use `data-testid` here instead of `getByRole('button')`?

packages/react-kit/src/shared/components/Button/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const textClasses: Record<NonNullable<ButtonProps['action']>, string> = {
+  primary: 'text-white text-base',
+  secondary: 'text-gray-900 text-base',
+  secondaryNeutral: 'text-white text-base',
+}
+
+export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
+  text?: string
+  // iconName?: IconName
```
brtkx, line 24: I guess we want to keep the icon code and wait for the designs to see if we'll use it? Could we leave a TODO for why it's commented out please?

OmarBasem: Yeah it was commented out until `Icon` component is ready, added a todo

packages/react-kit/src/index.ts
OmarBasem: Thanks @SahilVasava, fixed!

packages/playground/package.json
OmarBasem: Refactored to `apps/react-example`

apps/react-example/src/App.tsx
```diff
@@ -0,0 +1,107 @@
...
+            Interactive demo of{' '}
+            <code className="text-sm bg-gray-100 px-1.5 py-0.5 rounded">
+              @zerodev/wallet-react-kit
+            </code>{' '}
+            components
+          </p>
+        </div>
+
+        {/* Button */}
+        <section className="bg-white rounded-2xl shadow-sm border border-gray-200 p-8 space-y-6">
```
brtkx: IMO we should only keep the Connect button here for now, not a list of all components. For testing components we should use storybook, and the example app should just show what you can build with all the components.

packages/react-kit/.storybook/main.ts
```diff
@@ -0,0 +1,22 @@
+import type { StorybookConfig } from '@storybook/react-vite'
+
+import { dirname } from 'path'
+
+import { fileURLToPath } from 'url'
+
+/**
```
brtkx: We're not using PnP and storybook lives in a single package. Thinking if we need this at all, and if we can simplify this to:
```typescript
 const config: StorybookConfig = {
    stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
    addons: [
      '@storybook/addon-a11y',
      '@storybook/addon-docs',
    ],
    framework: '@storybook/react-vite',
  }
  export default config
```

packages/react-kit/src/auth/components/CodeInput/CodeInput.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const meta = {
+  title: 'Auth/CodeInput',
+  component: CodeInput,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs'],
+  argTypes: {
+    length: { control: { type: 'number', min: 4, max: 8 } },
```
brtkx, line 13: This doesn't seem to work right now. If I change length in storybook, it's always 6 regardless.

packages/react-kit/src/auth/components/CodeInput/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+    inputRefs.current[index]?.focus()
+  }
+
+  useEffect(() => {
+    if (autoFocus) {
+      inputRefs.current[0]?.focus()
+    }
+  }, [autoFocus])
+
+  const handleChange = (index: number, raw: string) => {
```
brtkx, line 43: I'm not able to replace a digit. If a digit is already inserted, I focus it and try to insert a different digit - it doesn't let me.

packages/react-kit/src/shared/components/Button/Button.stories.tsx
brtkx, line 1: nit: button texts in storybook don't have any padding x. I don't mind it being done when we get the designs, but just flagging.

packages/react-kit/package.json
```diff
@@ -0,0 +1,85 @@
...
+  },
+  "keywords": [
+    "react",
+    "components",
+    "ui",
+    "wallet",
+    "zerodev",
+    "web3"
+  ],
+  "license": "MIT",
```
brtkx: waiting for legal

packages/react-kit/README.md
```diff
@@ -0,0 +1,39 @@
...
+pnpm build
+
+# Watch mode
+pnpm dev
+
+# Type check
+pnpm typecheck
+```
+
+## License
```
brtkx: waiting for legal

brtkx (changes requested):

brtkx (approved):

PR #71, opened 2026-04-01
feat(core): rename stampers and expose them on createZeroDevWallet

andrascodes:

* **#142**
* **#141**
* **#131**
* **#80**
* **#124**
* **#76**
* **#75**
* **#74**
* **#73**
* **#72**
* ➡️ **#71**
* `main`
</details></td></tr></table>

Closes FS-1925, FS-1946

Renamed the stampers to a more generic name `apiKeyStamper` and `passkeyStampers`. Exposed them so they can be passed in through `createZeroDevWallet()`. But I haven't removed the default `indexedDbStamper` and `webauthnStamper` so this is not a breaking change.

(will add more documentation later)

### DIFF
--- a/packages/core/src/core/createZeroDevWallet.ts
+++ b/packages/core/src/core/createZeroDevWallet.ts
@@ -205,7 +205,6 @@ export async function createZeroDevWallet(
           organizationId: parsedSession.organizationId,
           stamperType: 'indexedDb',
           sessionType: SessionType.READ_WRITE,
-          publicKey: compressedPublicKeyHex,
           token: data.session,
           expiry: parsedSession.expiry,
           createdAt: Date.now(),
@@ -230,7 +229,6 @@ export async function createZeroDevWallet(
           if (data.session) {
             // Parse the JWT to get session data
             const parsedSession = parseSession(data.session)
-            const publicKey = await client.apiKeyStamper.getPublicKey()
             const session: ZeroDevWalletSession = {
               id: `session_oauth_${Date.now()}`,
               userId: parsedSession.userId,
@@ -240,7 +238,6 @@ export async function createZeroDevWallet(
               token: data.session,
               expiry: parsedSession.expiry,
               createdAt: Date.now(),
-              publicKey: publicKey || '',
             }
             await sessionStorageManager.storeSession(session, session.id)
           }
@@ -289,7 +286,6 @@ export async function createZeroDevWallet(
               expiry:
                 Date.now() +
                 Number(DEFAULT_SESSION_EXPIRATION_IN_SECONDS) * 1000,
-              publicKey: compressedPublicKeyHex,
               token: loginData.session,
             }
             await sessionStorageManager.storeSession(session, session.id)
@@ -322,7 +318,6 @@ export async function createZeroDevWallet(
               expiry:
                 Date.now() +
                 Number(DEFAULT_SESSION_EXPIRATION_IN_SECONDS) * 1000,
-              publicKey: generatedPublicKey,
               token: loginData.session,
             }
             await sessionStorageManager.storeSession(session, session.id)
@@ -436,7 +431,6 @@ export async function createZeroDevWallet(
                 token: data.session,
                 expiry: parsedSession.expiry,
                 createdAt: Date.now(),
-                publicKey: parsedSession.publicKey,
               }
               await sessionStorageManager.storeSession(session, session.id)
             }
--- a/packages/core/src/core/createZeroDevWallet.ts
+++ b/packages/core/src/core/createZeroDevWallet.ts
@@ -188,7 +188,7 @@ export async function createZeroDevWallet(
       if (!activeSession) {
         throw new Error('No active session')
       }
-      if (activeSession.stamperType === 'indexedDb') {
+      if (activeSession.stamperType === 'apiKey') {
         const compressedPublicKeyHex =
           await client.apiKeyStamper.prepareKeyRotation()
         const data = await client.loginWithStamp({
@@ -203,7 +203,7 @@ export async function createZeroDevWallet(
           id: `session_indexedDb_${Date.now()}`,
           userId: parsedSession.userId,
           organizationId: parsedSession.organizationId,
-          stamperType: 'indexedDb',
+          stamperType: 'apiKey',
           sessionType: SessionType.READ_WRITE,
           token: data.session,
           expiry: parsedSession.expiry,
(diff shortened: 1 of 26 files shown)


packages/core/src/utils/utils.ts
```diff
@@ -37,6 +37,7 @@ export function parseSession(
     organizationId,
     expiry: exp,
     token: publicKey,
+    publicKey,
```
OmarBasem: Why have 2 fields returning the same value?

andrascodes: Thanks! 🙏 I forgot to remove the 2nd one but I will do that now.

packages/core/src/core/createZeroDevWallet.ts
SahilVasava: Non blocking comment. But tightened type `publicKey: string` from `publicKey?: string` [here](https://github.com/OffchainLabs/zerodev-wallet-sdk/pull/71/changes#diff-17d42ee5e9e12c3fd1ec97a932a5729866ecc4475e66fc903bf1bfdf7ff709f3L15) made me realize we prob should have strict non null check for `publicKey` and fail faster instead of assigning empty string `''`. Logically I don't see any bugs, but just for future safety.

packages/core/src/types/session.ts
```diff
@@ -3,6 +3,7 @@ export enum SessionType {
   READ_WRITE = 'SESSION_TYPE_READ_WRITE',
 }
 
+// TODO: Change `stamperType` to `"apiKey" | "passkey"`
 export type StamperType = 'iframe' | 'indexedDb' | 'passkey'
```
SahilVasava: We should rename stamper type all atomically in same PR so that there are no dangling inconsistencies.

packages/core/src/utils/exportPrivateKey.ts
```diff
@@ -71,8 +71,9 @@ export async function exportPrivateKey(
     },
   })
 
+  // TODO: Change `stamperType` to `"apiKey" | "passkey"`
```
SahilVasava: Same here. Better to resolve it in same PR.

packages/core/src/utils/exportWallet.ts
```diff
@@ -56,9 +56,8 @@ export async function exportWallet(
 
     const listWalletsStamp =
       await wallet.client[
-        session.stamperType === 'indexedDb'
-          ? 'indexedDbStamper'
-          : 'webauthnStamper'
+        // TODO: Change `stamperType` to `"apiKey" | "passkey"`
+        session.stamperType === 'indexedDb' ? 'apiKeyStamper' : 'passkeyStamper'
```
SahilVasava: Same here. Better to resolve it in same PR.

packages/core/src/utils/exportWallet.ts
```diff
@@ -93,9 +92,8 @@ export async function exportWallet(
     })
     const exportWalletStamp =
       await wallet.client[
-        session.stamperType === 'indexedDb'
-          ? 'indexedDbStamper'
-          : 'webauthnStamper'
+        // TODO: Change `stamperType` to `"apiKey" | "passkey"`
+        session.stamperType === 'indexedDb' ? 'apiKeyStamper' : 'passkeyStamper'
```
SahilVasava: Same here.

packages/core/src/core/createZeroDevWallet.ts
```diff
@@ -181,30 +179,23 @@ export async function createZeroDevWallet(
         throw new Error('No active session')
       }
       if (activeSession.stamperType === 'indexedDb') {
```
SahilVasava: Same here.

packages/core/src/core/createZeroDevWallet.ts
```diff
@@ -181,30 +179,23 @@ export async function createZeroDevWallet(
...
+          stampWith: 'apiKey',
         })
-        await client.indexedDbStamper.resetKeyPair(newKeyPair)
+        await client.apiKeyStamper.commitKeyRotation()
         const parsedSession = parseSession(data.session)
         const session: ZeroDevWalletSession = {
           id: `session_indexedDb_${Date.now()}`,
           userId: parsedSession.userId,
           organizationId: parsedSession.organizationId,
           stamperType: 'indexedDb',
```
SahilVasava: Same here.

packages/core/src/core/createZeroDevWallet.ts
SahilVasava: Same here.

packages/core/src/stampers/types.ts
```diff
@@ -30,7 +30,33 @@ export type IframeStamper = Stamper & {
   applySettings(settings: { styles?: Record<string, string> }): Promise<boolean>
 }
 
-export type IndexedDbStamper = Stamper & {
-  resetKeyPair: (externalKeyPair?: CryptoKeyPair) => Promise<void>
+export type ApiKeyStamper = Stamper & {
+  /** Generate + activate a new key pair immediately (simple cases: login init, logout). */
+  resetKeyPair: () => Promise<void>
+  /** Generate a new key pair internally, return its public key, but keep the OLD key active for stamp(). */
```
SahilVasava: NIT: 
docstring should say: `compressed public key` for the sake of AI agents and humans! ^_^

packages/core/src/utils/utils.ts
SahilVasava: Heads up — Andras, the line to drop is line 40 (`token: publicKey`), not line 41. Line 41 (`publicKey,`) is actually used: the OTP block at `createZeroDevWallet.ts:424` reads `parsedSession.publicKey` to construct the session. Line 40 is the dead/misnamed one — none of the 5 call sites of `parseSession()` read `parsedSession.token`, they all set `token: data.session` (the actual JWT) directly. Only `utils.test.ts:37` asserts on `.token`, so that test line needs updating too.

Final shape of the return:
```ts
return {
  sessionType,
  userId,
  organizationId,
  expiry: exp,
  publicKey,
}
```

SahilVasava (commented):
Reviewed. The architectural direction is right — moving stampers to injected dependencies and the prepare/commit key rotation API is a clear improvement. Comments are mostly about consistency follow-through:

- The `StamperType` rename is half-finished; doing it atomically here avoids the `// TODO: @stamper-type` markers landing in main.
- `publicKey` field tightening surfaced a pre-existing weakness in the OAuth path (non-blocking).
- Misc nits on docstrings.

Approve conceptually, just want the rename consistency before merge.

SahilVasava (approved):
LGTM — all prior review comments addressed (StamperType renamed to 'apiKey' | 'passkey', OAuth path now fails loud on missing public key, parseSession deduped to single token field, docstrings clarified).

PR #72, opened 2026-04-01
refactor(core): use atob and btoa instead of Buffer

andrascodes:

* **#142**
* **#141**
* **#131**
* **#80**
* **#124**
* **#76**
* **#75**
* **#74**
* **#73**
* ➡️ **#72**
* `main`
</details></td></tr></table>

Closes FS-1926

Removing the usage of Buffer and switching to `atob` and `btoa` + using custom code. It works but this is probably best to replace with something more robust.

### DIFF
--- a/packages/core/src/utils/utils.ts
+++ b/packages/core/src/utils/utils.ts
@@ -18,7 +18,8 @@ export function parseSession(
     throw new Error('Invalid JWT: Missing payload')
   }
 
-  const decoded = JSON.parse(Buffer.from(payload, 'base64').toString())
+  const base64 = payload.replace(/-/g, '+').replace(/_/g, '/')
+  const decoded = JSON.parse(atob(base64))
   const {
     exp,
     public_key: publicKey,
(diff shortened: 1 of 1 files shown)


OmarBasem (commented):
Can you add context on the PR description on why we are removing `Buffer`, is it to make it work with RN?

andrascodes:
> Can you add context on the PR description on why we are removing `Buffer`, is it to make it work with RN?

Yes. No Buffer in Hermes. Do you have a better alternative solution for this? I'd be happy to change it.

SahilVasava (dismissed):
LGTM. Correctly addresses the base64url issue by doing the explicit `-`→`+`, `_`→`/` conversion before `atob`. The `Uint8Array` + `String.fromCharCode` + `btoa` round-trip is the standard browser-safe pattern.

brtkx (approved):

SahilVasava (approved):
LGTM (re-approving after force-push). atob/btoa swap is correct; base64url ↔ base64 conversion handled symmetrically.

PR #73, opened 2026-04-01
feat(react): expose stamper, storage and shouldInit params

andrascodes:

* **#168**
* **#142**
* **#141**
* **#131**
* **#80**
* **#124**
* **#76**
* **#75**
* **#74**
* ➡️ **#73**
* `main`
</details></td></tr></table>

Closes FS-1927

Passing through the new `wallet-core` options (storage, stampers...etc.) to `wallet-react` as well.
Additionally, removing or guarding the usages of `window` in `wallet-react` - added a `shouldInit` param to replace the check for "only init client-side" (still defaulting to `typeof window !== "undefined"` for now)

### DIFF
--- a/packages/react/src/connector.ts
+++ b/packages/react/src/connector.ts
@@ -6,12 +6,16 @@ import {
   createZeroDevPaymasterClient,
 } from '@zerodev/sdk'
 import { getEntryPoint, KERNEL_V3_3 } from '@zerodev/sdk/constants'
-import type { StorageAdapter } from '@zerodev/wallet-core'
+import type {
+  ApiKeyStamper,
+  PasskeyStamper,
+  StorageAdapter,
+} from '@zerodev/wallet-core'
 import { createZeroDevWallet, KMS_SERVER_URL } from '@zerodev/wallet-core'
 import { type Chain, createPublicClient, createWalletClient, http } from 'viem'
 import { handleOAuthCallback, type OAuthProvider } from './oauth.js'
 import { createProvider } from './provider.js'
-import { createZeroDevWalletStore } from './store.js'
+import { type CreateStoreOptions, createZeroDevWalletStore } from './store.js'
 import { getAAUrl } from './utils/aaUtils.js'
 
 // OAuth URL parameter used to detect callback
@@ -28,7 +32,7 @@ async function detectAndHandleOAuthCallback(
   wallet: Awaited<ReturnType<typeof createZeroDevWallet>>,
   store: ReturnType<typeof createZeroDevWalletStore>,
 ): Promise<boolean> {
-  if (typeof window === 'undefined') return false
+  if (typeof window === 'undefined' || !window.location?.search) return false
 
   const params = new URLSearchParams(window.location.search)
   const isOAuthCallback = params.get(OAUTH_SUCCESS_PARAM) === 'true'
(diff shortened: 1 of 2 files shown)


OmarBasem:
> Passing through the new wallet-core options (storage, stampers...etc.) to wallet-react as well.
Additionally, removing or guarding the usages of window in wallet-react - added a shouldInit param to replace the check for "only init client-side" (still defaulting to typeof window !== "undefined" for now)

Can you please add more details on not just what is being changed, but also why, for more clarity and context

packages/react/src/connector.ts
```diff
@@ -82,8 +86,14 @@ export type ZeroDevWalletConnectorParams = {
   chains: readonly Chain[]
   rpId?: string
   sessionStorage?: StorageAdapter
+  persistStorage?: CreateStoreOptions['storage']
+  apiKeyStamper?: Promise<ApiKeyStamper>
+  passkeyStamper?: Promise<PasskeyStamper>
```
SahilVasava: Forcing `Promise<ApiKeyStamper>` is a bit restrictive — synchronous stampers (e.g. test mocks, in-memory) have to be wrapped in `Promise.resolve(...)`. Suggest loosening to:

```ts
apiKeyStamper?: ApiKeyStamper | Promise<ApiKeyStamper>
passkeyStamper?: PasskeyStamper | Promise<PasskeyStamper>
```

`await` on a non-Promise value just returns the value, so the existing `if (params.apiKeyStamper) { apiKeyStamper = await params.apiKeyStamper }` already handles both. Zero behavior change, more flexibility for consumers.

packages/react/src/connector.ts
```diff
@@ -82,8 +86,14 @@ export type ZeroDevWalletConnectorParams = {
...
   autoRefreshSession?: boolean
   sessionWarningThreshold?: number
+  // Controls whether the connector should initialize
+  // On Web: pass `typeof window !== 'undefined'` to only initialize on client side
+  shouldInitialize?: boolean | (() => boolean)
```
SahilVasava: Naming nit: `shouldInitialize` reads imperatively but doesn't convey that it gates the *automatic* init that runs in `setup()` on connector mount. Suggest `autoInitialize` — clearer about what's being gated. Free to rename since this hasn't shipped yet.

```ts
autoInitialize?: boolean | (() => boolean)
```

SahilVasava (commented):
Substantive direction is right — exposing stamper, persistStorage, and shouldInitialize is exactly what RN consumers need. Three things inline:

- 🔴 Dup-store: the new configured store is overwritten by the existing unconfigured one; `persistStorage` is dead until #75. Fold the deletion in here.
- 🟡 Loosen `Promise<Stamper>` → `Stamper | Promise<Stamper>` so sync stampers don't need wrapping.
- ⚪ Naming nit: `shouldInitialize` → `autoInitialize`.

packages/react/src/connector.ts
```diff
@@ -136,9 +155,15 @@ export function zeroDevWallet(
...
+        ...(apiKeyStamper && { apiKeyStamper }),
+        ...(passkeyStamper && { passkeyStamper }),
       })
 
       // Create store
+      store = createZeroDevWalletStore({
+        ...(params.persistStorage && { storage: params.persistStorage }),
+      })
+
       store = createZeroDevWalletStore()
```
SahilVasava: 🔴 The new `store = createZeroDevWalletStore({...})` block above doesn't replace this line — both end up running, so the configured store is immediately overwritten by this unconfigured one. `params.persistStorage` is dead code in this PR. PR #75 fixes it by deleting this line, but the deletion belongs here so each PR is correct on its own.

Fix: delete this line.

SahilVasava (dismissed):
LGTM — resolutions look good: stamper params now accept sync or async (`ApiKeyStamper | Promise<ApiKeyStamper>`), `shouldInitialize` renamed to `autoInitialize` with clear semantics, and the duplicate `createZeroDevWalletStore` block is gone.

brtkx (approved):

PR #74, opened 2026-04-01
feat: add fetchOptions to core and react so Origin header can be sent on RN

andrascodes:

* **#168**
* **#142**
* **#141**
* **#131**
* **#80**
* **#124**
* **#76**
* **#75**
* ➡️ **#74**
* `main`
</details></td></tr></table>

Part of FS-1971

`fetch` doesn't automatically send `Origin` headers on React Native as there's no concept of a domain there. (It sends an IP instead.) So I solved this by passing through the `Origin` headers instead configurable via the `wallet-react` and `wallet-core` entrypoints.
We definitely have to come up with a better solution to this, but for the purpose of testing this works. I'm totally fine not merging this.

### DIFF
--- a/packages/react/src/connector.ts
+++ b/packages/react/src/connector.ts
@@ -8,6 +8,7 @@ import {
 import { getEntryPoint, KERNEL_V3_3 } from '@zerodev/sdk/constants'
 import type {
   ApiKeyStamper,
+  CreateTransportOptions,
   PasskeyStamper,
   StorageAdapter,
 } from '@zerodev/wallet-core'
@@ -104,6 +105,7 @@ export type ZeroDevWalletConnectorParams = {
   persistStorage?: CreateStoreOptions['storage']
   apiKeyStamper?: ApiKeyStamper | Promise<ApiKeyStamper>
   passkeyStamper?: PasskeyStamper | Promise<PasskeyStamper>
+  fetchOptions?: CreateTransportOptions['fetchOptions']
   autoRefreshSession?: boolean
   sessionWarningThreshold?: number
   /**
@@ -160,7 +162,10 @@ export function zeroDevWallet(
      *
      * No-ops if the chain is already set up.
      */
-    const setupChain = async (chainId: number) => {
+    const setupChain = async (
+      chainId: number,
+      httpOpts?: { fetchOptions?: CreateTransportOptions['fetchOptions'] },
+    ) => {
       const state = store.getState()
       const eoaAccount = state.eoaAccount
       if (!eoaAccount) throw new Error('Not authenticated')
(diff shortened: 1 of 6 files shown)


andrascodes:
I'll actually change this so we pass in an `appId` only. Then inside the SDK I'll just create the Origin header from the `appId`. I think that will match better what the KMS is planning to accept.

packages/react/src/connector.ts
```diff
@@ -89,6 +89,7 @@ export type ZeroDevWalletConnectorParams = {
   persistStorage?: CreateStoreOptions['storage']
   apiKeyStamper?: Promise<ApiKeyStamper>
   passkeyStamper?: Promise<PasskeyStamper>
+  fetchOptions?: Omit<RequestInit, 'body' | 'method' | 'signal'>
```
SahilVasava: Type duplication nit: this inline `Omit<RequestInit, 'body' | 'method' | 'signal'>` is the same shape as `CreateTransportOptions['fetchOptions']` over in core. `createZeroDevWallet.ts` already reuses that via indexed access:

```ts
fetchOptions?: CreateTransportOptions['fetchOptions']
```

Suggest the same here so the type stays in sync if the omit list ever changes:

```ts
import type { ApiKeyStamper, CreateTransportOptions, PasskeyStamper, StorageAdapter } from '@zerodev/wallet-core'

// ...
fetchOptions?: CreateTransportOptions['fetchOptions']
```

Indexed access preserves the Omit transparently — same effective type, single source of truth.

packages/react/src/connector.ts
```diff
@@ -277,13 +279,15 @@ export function zeroDevWallet(
             account: kernelAccount,
             bundlerTransport: http(
               getAAUrl(params.projectId, activeChainId, params.aaUrl),
+              { fetchOptions: params.fetchOptions },
```
SahilVasava: `{ fetchOptions: params.fetchOptions }` is repeated four times across this function (here + line ~288 paymaster, ~392 bundler-on-chain-switch, ~398 paymaster-on-chain-switch). If the shape ever changes or a fifth site is added, easy to drift.

Hoist once at the top:

```ts
const httpOpts = params.fetchOptions ? { fetchOptions: params.fetchOptions } : undefined
// ...
bundlerTransport: http(getAAUrl(...), httpOpts),
```

Non-blocking, just cleanup.

SahilVasava (commented):
Direction is right — `fetchOptions` is the right plumbing for RN's Origin-header use case, and the `Omit<RequestInit, 'body' | 'method' | 'signal'>` is the correct restriction (SDK builds body, sets method, controls signal).

Two cleanup suggestions inline:
- Reuse `CreateTransportOptions['fetchOptions']` instead of duplicating the Omit
- Hoist `{ fetchOptions: params.fetchOptions }` once instead of repeating it four times

(Out of scope: the user's `fetchOptions.headers` can override SDK's `Content-Type` due to merge order in `rest.ts:52-56`. Pre-existing, worth a follow-up but doesn't block this.)

SahilVasava (dismissed):
LGTM — type duplication removed (uses `CreateTransportOptions['fetchOptions']`) and the repeated `{ fetchOptions: params.fetchOptions }` is now a single `httpOpts` local.

brtkx (approved):

PR #76, opened 2026-04-01
feat: add expo-example app

andrascodes:

* **#168**
* **#142**
* **#141**
* **#131**
* **#80**
* **#124**
* ➡️ **#76**
* `main`
</details></td></tr></table>

Closes FS-1929

Added an `expo-example` app to test `wallet-react`. I needed to do a couple stuff to make Expo/Metro work with the monorepo and the packages:
1. [Adding](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/60e085c8581cc61e4d44548e8399db957aee6762/apps/expo-example/babel.config.js) `babel.config.js` to get rid of the `import.meta` error

<del>Added `nodeLinker: hoisted` to the [workspace file](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/60e085c8581cc61e4d44548e8399db957aee6762/pnpm-workspace.yaml) and the [NPM config file](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/60e085c8581cc61e4d44548e8399db957aee6762/.npmrc).
    - This is needed to make the example app work, however, it can cause drawbacks with module resolution. ([Source](https://docs.expo.dev/guides/monorepos/#package-managers-with-isolated-dependencies)) We can discuss the best strategy here. On one hand, the example apps make it a lot easier to test the packages while developing. It's [not a good option to install local modules in Expo](https://github.com/expo/expo/issues/16239) as it's cumbersome and requires frequent bundle cache and `node_modules` resets. On the other hand, I do understand the drawbacks can be more annoying.</del>

<del>Hoisted deps alone did not work to resolve multiple package versions so I had to add overrides [here](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/60e085c8581cc61e4d44548e8399db957aee6762/package.json#L58).</del>

Got inspired from [Omar's PR](https://github.com/OffchainLabs/zerodev-wallet-sdk/pull/81) and I added a Metro config instead to link to the correct modules. This might make some new package installs brittle so we need to keep extending it but it beats having to make global dep management modifications to the repo for the sake of an example app.
Claude's explanation:

> ⏺ The metro config solves a problem caused by pnpm in monorepos: duplicate package copies.
> 
>   When packages/react and apps/expo-example both depend on wagmi, pnpm installs two separate copies in its .pnpm store (even with node-linker=hoisted) because their transitive dependency trees differ slightly. Metro then bundles both copies, meaning WagmiProvider from one copy creates a React context that useConfig from the other copy can't find.
> 
>   What the config does:
> 
>   config.resolver.resolveRequest = (context, moduleName, platform) => {
> 
>   It intercepts Metro's module resolution. For every import in the bundle, this function decides where the module comes from.
> 
>   Two categories of overrides:
> 
>   1. workspaceSingletons (react, wagmi, @wagmi/core, @tanstack/react-query, react-dom) — Only redirected when the import originates from packages/* (your workspace libraries). This is scoped because redirecting these globally would break wagmi's own internal imports of @wagmi/core etc.
>   2. globalSingletons (tslib) — Redirected for ALL imports regardless of origin. Safe because tslib is a standalone utility with no internal dependencies.
> 
>   How the redirect works:
> 
>   return context.resolveRequest(
>     { ...context, originModulePath: appPackageJson },
>     moduleName,
>     platform,
>   )
> 
>   It lies to Metro about where the import is coming from — instead of resolving from packages/react/dist/..., it resolves as if the import came from apps/expo-example/package.json. Metro then walks up from there, finds apps/expo-example/node_modules/wagmi, and uses the app's copy. This ensures a single copy of each package ends up in the bundle.


apps/expo-example/lib/passkeyStamper.ts
```diff
@@ -0,0 +1,43 @@
...
+  createPasskey,
+  PasskeyStamper as TurnkeyPasskeyStamper,
+} from '@turnkey/react-native-passkey-stamper'
+import type {
+  PasskeyRegistrationOptions,
+  PasskeyStamper,
+} from '@zerodev/wallet-core'
+import { v4 as uuidv4 } from 'uuid'
+
+export async function createPasskeyStamper({
```
SahilVasava: Naming convention nit (Option A): rest of the codebase names stamper factories after their *implementation* and types after the *capability* (`createIndexedDbStamper(): ApiKeyStamper`, `createWebauthnStamper(): PasskeyStamper`, `createSecureStoreStamper(): ApiKeyStamper`). Suggest renaming this to `createReactNativePasskeyStamper` (and the file to `reactNativePasskeyStamper.ts`) so it doesn't read like the canonical passkey factory — once we add an iOS-specific or test passkey stamper, plain `createPasskeyStamper` becomes ambiguous.

apps/expo-example/lib/passkeyStamper.ts
```diff
@@ -0,0 +1,43 @@
...
+  rpId,
+}: {
+  rpId: string
+}): Promise<PasskeyStamper> {
+  const inner = new TurnkeyPasskeyStamper({ rpId })
+
+  return {
+    async getPublicKey() {
```
SahilVasava: `getPublicKey()` silently returns `null` with a commented-out call to the inner stamper. Two issues:

1. The commented `inner.()` line should be deleted (looks like work-in-progress).
2. Returning `null` is misleading — callers checking `if (!publicKey)` will see this as "no key set" when really it's "this stamper doesn't have a stable public key concept."

Suggest making it explicit:

```ts
async getPublicKey() {
  throw new Error('Passkey stamper does not expose a stable public key — use credentialId from registration')
}
```

That way the failure has a useful stack trace if someone reaches this code path by mistake.

pnpm-workspace.yaml
```diff
@@ -1,11 +1,8 @@
...
+  - apps/*
 
-# Security: Only install packages that are at least 24 hours old (1440 minutes)
-# This helps protect against typosquatting and malicious packages
 minimumReleaseAge: 1440
 
-# Exclude trusted packages from age restriction
 minimumReleaseAgeExclude:
-  - '@zerodev/*'
+  - '@zerodev/*'
```
SahilVasava: The `minimumReleaseAge: 1440` setting survives but the security-rationale comments above it are dropped:

```yaml
# Security: Only install packages that are at least 24 hours old (1440 minutes)
# This helps protect against typosquatting and malicious packages
minimumReleaseAge: 1440

# Exclude trusted packages from age restriction
minimumReleaseAgeExclude:
```

Without the *why*, next person looking at this is more likely to remove the setting as "random number, why is this here?" Also lost the trailing newline (most linters flag this), and quote-stripping is inconsistent — `'packages/*'` → `packages/*` but `'@zerodev/*'` kept its quotes.

Suggest restoring the comments and newline.

SahilVasava (commented):
Looks solid. Three inline nits — naming, `getPublicKey` returning null, and lost yaml comments. Nothing blocking.

SahilVasava (dismissed):
LGTM — all three nits addressed: factory renamed to `createReactNativePasskeyStamper` (Option A), `getPublicKey` issue gone (PasskeyStamper type doesn't require it), and the pnpm-workspace.yaml security-rationale comments are back.

brtkx (approved):

PR #78, opened 2026-04-02
feat: create `react-native-kit`

OmarBasem:
Closes FS-1912

### Summary

This PR sets up and creates `react-native-kit` package, and adds a `react-native-example` app (including `react-native-web`)

I will add storybook for react-native ~in a follow-up PR~  as it requires some extra setup
Can add it later whenever we need UI support on React Native

Just added one `Button` component to test the setup


andrascodes:
@OmarBasem Would it be possible for you to split out `react-native-kit` into its own PR (or make this PR only for `react-native-kit`)? I already added an `expo-example` app [here](https://github.com/OffchainLabs/zerodev-wallet-sdk/pull/76). Do you think it's possible for you to extend that?
The `react-native-kit` setup I'd like to merge before my stack, because I'd like to move some files (`asyncSessionStorage`, `passkeyStamper, `secureStoreStamper`) over there and remove from the `expo-example`. It's possible also my stack would still be in review for a while and most of what's left for me to do is platform specific work so I'd probably make new PRs on top of the stack extending `react-native-kit`. That's why I'd want that to be a separate PR.

OmarBasem:
@andrascodes no problem I can split it

brtkx (dismissed):

andrascodes (dismissed):
thx!!!

brtkx (approved):

PR #80, opened 2026-04-02
feat(react): enable OAuth in React Native

andrascodes:

* **#168**
* **#142**
* **#141**
* **#131**
* ➡️ **#80**
* `main`
</details></td></tr></table>

Closes FS-1922, FS-2000

## This PR comes after #124. So make sure you review #124 before this one if you want to go in a chronological order.

**Update: Changed this PR to be on top of the stack and added Passkeys and Android setup before this. This way this PR can be tested by just running `pnpm android` and using the Android Emulator.**

Refactored the OAuth implementation in `wallet-react` to be more contained in a single place so it's easier to create web and mobile modules for it.
The new structure of the `authenticateOAuth` action looks like this:
```
// setup - same as it was
// `getSessionId` - injected module for web and mobile, this takes care of opening the popup and waiting for the sessionId, sessionId is the result of a Promise
// `completeAuth` - calling `wallet.auth` and changing the state, this used to be here too so basically unchanged
```
The main change is the behavior of the web OAuth flow. The new code doesn't rely on using `postMessage` and a message listener anymore. It instead polls for the URL change of the popup. The main advantage of this is that we don't need any of the additional code that we had in the old logic where the popup had to open, load the app, run JS to detect the OAuth callback and complete the auth with the `sessionId`. We can do that all in the `getSessionIdWeb` module. This meant that we could get rid of the OAuth related code in `connector.ts`.
So the new logic just checks the URL of the popup every 250ms and if it's on the same origin (same URL as the app) then it tries to get the sessionId param off of the query params.
An additional security advantage to this is that checking for same origin is implicit as we can't get the popup's URL across origins. So if the popup is still on the Google OAuth flow the check errors on `popup.location.href` already. In the old flow we had to make sure to check for `origin === expectedOrigin` to be secure.


packages/react/src/authenticateOAuth.ts
```diff
@@ -60,85 +50,96 @@ export function buildBackendOAuthUrl(params: BackendOAuthFlowParams): string {
...
-    if (!event.data || typeof event.data !== 'object') return
-
-    if (event.data.type === 'oauth_success') {
-      cleanup()
-      onSuccess(event.data.sessionId || '')
-    } else if (event.data.type === 'oauth_error') {
-      cleanup()
-      onError(new Error(event.data.error || 'OAuth authentication failed'))
-    }
+export async function authenticateOAuth(
```
SahilVasava, line 56: Ideally all actions should have separate files. For this PR it makes sense to keep `authenticateOAuth` action flat in `actions.ts` like before. Separate files could be tackled in another refactor PR.

packages/react/src/getSessionIdWeb.ts
```diff
@@ -0,0 +1,70 @@
...
+  const popup = openOAuthPopup(oauthUrl)
+  if (!popup) throw new Error('Failed to open OAuth login window')
+  return pollPopupForSession(popup, expectedOrigin)
+}
+
+function pollPopupForSession(
+  popup: Window,
+  expectedOrigin: string,
+): Promise<string> {
+  return new Promise((resolve, reject) => {
```
SahilVasava, line 38: No timeout. If the popup never reaches `expectedOrigin` (backend bug, infinite redirect, network drop), this Promise hangs forever — the interval keeps polling, the popup stays open, the React Query mutation never settles, the consumer's UI is stuck.

Suggest a sensible default with override:

```ts
export async function getSessionIdWeb(
  oauthUrl: string,
  expectedOrigin: string,
  timeoutMs = 5 * 60 * 1000, // 5min
): Promise<string> {
  // ...
}
```

In `pollPopupForSession`, race against `setTimeout(() => reject(new Error('OAuth timed out')), timeoutMs)` and clear it on resolve/reject.

packages/react/src/getSessionIdWeb.ts
```diff
@@ -0,0 +1,70 @@
...
+    const tick = setInterval(() => {
+      if (popup.closed) {
```
SahilVasava: Race: this `popup.closed` check runs *before* the URL check. If the success-redirect lands and the popup auto-closes (via `popup.close()` below) within the same 250ms tick, the next iteration sees `closed=true` first and rejects with "OAuth popup was closed" — even though auth succeeded.

Reverse the order so URL is read first, fall back to closed:

```ts
const tick = setInterval(() => {
  try {
    const url = new URL(popup.location.href)
    if (url.origin === expectedOrigin) {
      // ... existing URL-based resolve/reject
      return
    }
  } catch {
    // cross-origin, keep polling
  }
  if (popup.closed) {
    clearInterval(tick)
    reject(new Error('OAuth popup was closed'))
  }
}, 250)
```

apps/expo-example/oauth/createNativeOAuthGetSessionId.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (!sid) throw new Error('OAuth redirect missing session_id')
+      return sid
+    })
+
+    try {
+      const sid = await Promise.race([fromDeepLink, fromBrowser])
+      // Close the auth tab if the deep-link branch won — the OS brought
+      // the app to foreground but left the Custom Tab in the back stack.
+      // No-op when fromBrowser already won.
+      WebBrowser.dismissBrowser()
```
SahilVasava: `WebBrowser.dismissBrowser()` only fires inside the `try` success block. If `fromDeepLink` rejects (error param in URL) or `fromBrowser` throws (user cancels, network error), `Promise.race` rejects fast — but the auth tab on Android Custom Tabs / iOS ASWebAuth stays open until the user dismisses manually.

Move to `finally` alongside `sub.remove()`:

```ts
try {
  return await Promise.race([fromDeepLink, fromBrowser])
} finally {
  WebBrowser.dismissBrowser()
  sub.remove()
}
```

packages/react/src/authenticateOAuth.ts
```diff
@@ -60,85 +50,96 @@ export function buildBackendOAuthUrl(params: BackendOAuthFlowParams): string {
...
-    clearInterval(checkWindowClosed)
+  const publicKey = await wallet.getPublicKey()
+  if (!publicKey) {
+    throw new Error('Failed to get wallet public key')
   }
 
-  window.addEventListener('message', handleMessage)
+  const returnTo = redirectUri
+    ? `${redirectUri}?oauth_success=true&oauth_provider=${provider}`
+    : `${window.location.origin}?oauth_success=true&oauth_provider=${provider}`
```
SahilVasava: Same `returnTo` path-drop bug we just fixed in `alpha.19` (#143): `${window.location.origin}` strips the parent's pathname, so the popup lands on `/` not whatever route invoked the OAuth.

Less harmful here than in alpha.18 because the new design polls `popup.location.href` from the parent (works regardless of what page the popup loads), but the user *does* see the wrong page flash in the popup briefly before it closes. UX nit.

Please cherry-pick the alpha.19 fix when rebasing this onto main — using `window.location.href` as the base preserves the pathname:

```ts
const returnUrl = new URL(redirectUri ?? window.location.href)
returnUrl.hash = ''
returnUrl.searchParams.set('oauth_success', 'true')
returnUrl.searchParams.set('oauth_provider', provider)
```

packages/react/src/hooks/useAuthenticateOAuth.ts
```diff
@@ -6,43 +6,60 @@ import {
...
     },
     mutationKey: ['authenticateOAuth'],
   })
 }
 
 export declare namespace useAuthenticateOAuth {
+  type Variables = {
+    provider: authenticateOAuth.Parameters['provider']
+    connector?: authenticateOAuth.Parameters['connector']
+  }
```
SahilVasava: This is the only hook in the package with a separate `Variables` type — every other hook (`useLoginPasskey`, `useSendOTP`, etc.) uses `actionName.Parameters` directly as the mutation variables.

The split was forced because `authenticateOAuth.Parameters` mixes hook-level config (`getSessionId`, `redirectUri` — set once when the hook is created) with call-level args (`provider` — passed to `mutate()`). Cleanest fix: drop `getSessionId`/`redirectUri` from the action's `Parameters`, keep them only at the hook level, pass them to the action invocation inside `mutationFn`:

```ts
// authenticateOAuth.Parameters becomes just
{ provider: OAuthProvider, connector?: Connector }

// hook config bag stays the same
useAuthenticateOAuth({ getSessionId, redirectUri })

// no Variables type — uses authenticateOAuth.Parameters like all other hooks
```

This brings the hook's shape in line with the rest and removes a one-off pattern.

SahilVasava (commented):
Architecture is right — pluggable `getSessionId` is a much cleaner native story than what the old design could have grown into. Five inline notes:

- 🟡 No timeout on `getSessionIdWeb` — promise hangs on stuck popups
- 🟡 `popup.closed` checked before URL — closed-vs-success race
- 🟡 Native `dismissBrowser()` only on success — auth tab stays open on error
- 🟡 Same `returnTo` path-drop as #143 fixed — cherry-pick when rebasing
- 🟡 `Variables` type only in this hook — drop by aligning action's `Parameters`

Nothing blocking the architectural direction; all addressable in this PR or follow-ups.

packages/react/src/getSessionIdWeb.ts
andrascodes: Good catch on the reorder — applied. One correction on the mechanism though: the trigger isn't our own
  popup.close() "below". clearInterval(tick) is synchronous and runs before popup.close(), so no further  
  poll can fire from that path; even if one did, the promise is already resolved.                       
                                                                                                          
  The real trigger is the OAuth callback page itself calling window.close() between polls. I verified this
   with a local repro (cross-origin provider → same-origin callback → window.close() on load): Chromium   
  preserves popup.location.href after popup.closed flips to true, so a single tick can observe closed=true
   together with the success URL. Under the old check order, that tick rejected with "OAuth popup was     
  closed" despite auth having succeeded; with the reorder it resolves with the session id.

  Added a regression test (getSessionIdWeb.test.ts) that simulates the same-tick closed=true +            
  href=<success url> state — fails on the old order, passes on the new one.

SahilVasava (approved):
LGTM — all 6 nits addressed: `getSessionIdWeb` now takes `timeoutMs` (default 5min), URL is read before `popup.closed` check (with comment explaining the closed-vs-success race), native `dismissBrowser()` moved to `finally` so it fires on errors too, `returnUrl` uses `window.location.href` (path preserved), and `useAuthenticateOAuth` no longer has its own `Variables` type. Separate `authenticateOAuth.ts` file is fine given the adapter-pattern complexity.

PR #85, opened 2026-04-08
docs(react-kit): set up React Kit documentation

brtkx:
This will eventually move to https://docs.zerodev.app but for testing purposes we'll have it here, until React Kit is public.

Run `docs:dev` in `/packages/react-kit` to test.


packages/react-kit/docs/pages/react/features/transaction-signing.mdx
```diff
@@ -0,0 +1,36 @@
+import SignatureRequestExample from '../../../components/SignatureRequest'
```
OmarBasem, line 1: We need to add Auth under features also, right?

OmarBasem (approved):

packages/react-kit/docs/pages/react/features/transaction-signing.mdx
brtkx: yep, although we don't need to do it as part of https://github.com/OffchainLabs/zerodev-wallet-sdk/pull/88. We can have a follow up PR.

PR #88, opened 2026-04-09
feat(react-kit): Setup auth flow

OmarBasem:
Closes FS-1948

### Summary

This PR sets up the Auth flow, it adds the following:

- `useAuth` hook
- `AuthProvider`
- Different auth pages
- `AuthFlow.tsx` to handle the flow
- Auth types

Pages UI is just a rough UI as designs are not ready for development yet.

Added an example usage under react-example

### Demo

https://github.com/user-attachments/assets/52274980-5bfb-41f7-aec3-96b7fced26ed


packages/react-kit/src/auth/AuthProvider.tsx
brtkx, line 1: We don't need another provider for this. It could be part of the [ZeroDev wallet config](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/main/packages/react-kit/src/connector.ts#L21):
```typescript
export type ZeroDevKitConfig = {
  signing?: SigningConfig
  auth?: AuthConfig
}
```

Then we initialize the store in [zeroDevKitWallet](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/main/packages/react-kit/src/connector.ts#L43):
```typescript
if (params.config?.auth) {
  store.getState().setAuthConfig(params.config.auth)
}
```

That means we also don't need a separate auth store - it can be part of existing [reactKitStore](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/main/packages/react-kit/src/store.ts)  using a slice pattern:
```typescript
const createAuthSlice = (set, get) => ({
  config: null,
  setConfig: (config) => set({ config }),
  step: 'initializing',
  // ...
})

export type State = {
  auth: AuthSlice
  // signing is something I'd change on my end, for now can leave signing as is
  signing: SigningSlice
  userConfirmationListenerActive: boolean
  setUserConfirmationListenerActive: (active: boolean) => void
}
```

And then useAuth can access the store the same way [usePendingRequest.ts](https://github.com/OffchainLabs/zerodev-wallet-sdk/blob/main/packages/react-kit/src/signing/hooks/usePendingRequest.ts) does. In a later PR I'll extract a reusable useKitStore hook, but for now repeating the getStore helper is fine.

brtkx: Edit: it already uses `reactKitStore` my bad - lets just address the other stuff + add slice pattern to the store.

packages/react-kit/src/auth/authSlice.ts
```diff
@@ -0,0 +1,209 @@
+import type { StateCreator } from 'zustand'
+import type { AuthConfig, AuthMethod, AuthStep } from './types'
+
+export interface AuthSlice {
+  auth: {
+    // State
+    step: AuthStep
+    stepHistory: AuthStep[]
+    availableMethods: AuthMethod[]
```
brtkx: Could we rename it to `enabledMethods` so the naming is consistent with the wagmi config?

packages/react-kit/src/auth/authSlice.ts
brtkx, line 1: Can we call these `authStoreSlice` `AuthStoreSlice` `createAuthStoreSlice` just to avoid ambiguity if someone else was to look at the files and code

packages/react-kit/src/auth/authSlice.ts
```diff
@@ -0,0 +1,209 @@
...
+    availableMethods: [],
+    email: null,
+    otpId: null,
+    error: null,
+    pendingMethod: null,
+    resendAvailableAt: null,
+    config: null,
+
+    // Actions
+    initialize: async (config: AuthConfig) => {
```
brtkx: Don't need `async` here

packages/react-kit/src/connector.ts
```diff
@@ -46,12 +48,36 @@ export function zeroDevKitWallet(
   const baseFactory = zeroDevWallet(params)
   const store = createStore()
 
+  // Initialize auth config if provided
+  if (params.config?.auth) {
+    store
+      .getState()
+      .auth.initialize(params.config.auth)
+      .catch(() => {
+        // Initialization errors are handled internally
```
brtkx: I don't think we need those catch blocks?

packages/react-kit/src/auth/hooks/useAuth.ts
```diff
@@ -0,0 +1,140 @@
...
+    error,
+    isLoading: pendingMethod !== null,
+    pendingMethod,
+    availableMethods,
+    canResend: secondsUntilResend <= 0,
+    secondsUntilResend,
+    submitEmail,
+    submitOtp,
+    resendOtp,
+    selectMethod: useCallback(
```
brtkx: The store reference is stable so we shouldn't need any of these `useCallback`s

packages/react-kit/src/auth/hooks/useAuth.ts
```diff
@@ -0,0 +1,140 @@
...
+      (
+        method: Parameters<
```
brtkx: Wouldn't this be the same?
`method: AuthMethod`

packages/react-kit/src/auth/authSlice.ts
```diff
@@ -0,0 +1,209 @@
...
+      set((state) => ({
+        auth: {
+          ...state.auth,
+          config,
+          availableMethods: config.enabledMethods,
+          step: 'select-method',
+        },
+      }))
+    },
+
```
brtkx: I don't think we need to list all of different navigation functions here, such as `selectMethod`, `submitEmail`, etc. What if we did something like this:
```typescript
type StepAction =
  | { step: 'email-verification'; email: string }
  | { step: 'email-input' }
  | { step: 'otp-input' }
  | { step: 'verifying-otp' }
  // ...

goToStep: (action: StepAction) => {
  set((state) => ({
    auth: {
      ...state.auth,
      ...action,
      stepHistory: [...state.auth.stepHistory, state.auth.step],
    },
  }))
}
```

Then each component just does this:
```typescript
goToStep({ step: 'otp-input' })
goToStep({ step: 'email-verification', email })
```

I also don't think we need `pendingMethod` at all - from what I gathered it's only used for loading states, and each component should internally handle its loading state based on ZD hooks etc.

packages/react-kit/src/auth/authSlice.ts
```diff
@@ -0,0 +1,209 @@
...
+  auth: {
+    // State
+    step: AuthStep
+    stepHistory: AuthStep[]
+    availableMethods: AuthMethod[]
+    email: string | null
+    otpId: string | null
+    error: { message: string; recoverable: boolean } | null
+    pendingMethod: AuthMethod | null
+    resendAvailableAt: number | null
```
brtkx: Since we don't allow it to be set in the config, I don't think it needs to be part of the store. It could be handled inside the component and if later devs or product want this customizable, we can move things out.

For now I'd aim to keep this as lean as we can.

packages/react-kit/src/auth/hooks/useAuth.ts
```diff
@@ -0,0 +1,140 @@
...
+    } catch (err) {
+      store.getState().auth.setError({
+        message: err instanceof Error ? err.message : 'Failed to resend OTP',
+        recoverable: true,
+      })
+    }
+  }
+
+  // Derived: resend countdown
+  const [secondsUntilResend, setSecondsUntilResend] = useState(0)
```
brtkx: Does this and `secondsUntilResend` need to be part of this hook? I feel like it could be handled inside the component.

packages/react-kit/src/auth/pages/MethodPicker.tsx
```diff
@@ -0,0 +1,46 @@
...
+  passkey: 'Continue with Passkey',
+  'injected-wallet': 'Continue with Wallet',
+}
+
+export function MethodPicker({ showAll = false }: MethodPickerProps) {
+  const { availableMethods, selectMethod } = useAuth()
+
+  const displayMethods = showAll
+    ? availableMethods
+    : availableMethods.slice(0, 3)
```
brtkx: What's the reason for that?

OmarBasem: It was an outdated code, removed it, thanks

packages/react-kit/src/auth/authStoreSlice.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    goToStep: (action: StepAction) => {
+      const { auth } = get()
+      set((state) => ({
+        auth: {
+          ...state.auth,
+          step: action.step,
+          stepHistory: [...state.auth.stepHistory, auth.step],
+          // Merge any additional data from the action (like email)
+          ...('email' in action && { email: action.email }),
```
brtkx: This is just extra code for no benefit. Spreading `...action` does the same thing (and we don't have to add a new line for every new property later).

Neither approach is fully type safe - if we changed `{ step: 'email-verification'; email: string }` to `{ step: 'email-verification'; foo: string }` both would silently leak foo onto the store even though it doesn't exist there. This is something we should cover with tests in the next PR.

packages/react-kit/src/auth/authStoreSlice.ts
```diff
@@ -0,0 +1,141 @@
...
+        },
+      }))
+    },
+
+    selectMethod: (method: AuthMethod) => {
```
brtkx: I think we can simplify this as well. We could completely remove this part and in `MethodPicker.tsx` just do:
```typescript
const METHOD_STEPS: Record<AuthMethod, AuthStep> = {
  email: 'email-input',
  passkey: 'passkey-prompt',
  google: 'oauth-in-progress',
  'injected-wallet': 'wallet-selection',
}

<Button onClick={() => goToStep({ step: METHOD_STEPS[method] })} />
```

brtkx (approved):

PR #89, opened 2026-04-09
feat(react-kit): decode ETH and ERC-20 transfers in signing UI

brtkx:
Closes FS-1952

ETH and ERC-20 transactions are now decoded and have their own signing UI. You can test in the example app.


packages/react-kit/src/signing/components/SignatureRequest/index.tsx
brtkx, line 1: Moved from `/signing/components` to `/signing`

packages/react-kit/src/signing/components/SigningActions.tsx
```diff
@@ -0,0 +1,25 @@
+interface SigningActionsProps {
+  confirm: () => void
+  reject: () => void
+}
+
+export function SigningActions({ confirm, reject }: SigningActionsProps) {
+  return (
+    <div className="flex gap-3">
+      <button
```
OmarBasem: Why not use the `Button` component we have in shared

packages/react-kit/src/signing/index.tsx
```diff
@@ -0,0 +1,72 @@
+'use client'
```
OmarBasem, line 1: That file here is okay for now, but we can align on what to name and where to put the main file of a feature. For auth the main file exported is `AuthFlow.tsx` under `pages`. I think maybe a file named `index.tsx` directly under the feature `signing` or `auth` would make sense

packages/react-kit/src/signing/components/SigningActions.tsx
brtkx: I'll do that

packages/react-kit/src/signing/index.tsx
brtkx: We agreed async to put those files under signing/index.ts auth/index.ts etc

brtkx:
@OmarBasem I've addressed your comments

OmarBasem (approved):

PR #99, opened 2026-04-14
feat: support get-authenticators and remove get-email api

jayeshbhole:
Replaces get email api with get authenticators api.


SahilVasava (approved):
Lgtm!

PR #105, opened 2026-04-17
feat(react-kit): decode ERC-20 approvals in signing UI

brtkx:
Closes FS-1963

### DIFF
--- a/packages/react-kit/src/signing/index.tsx
+++ b/packages/react-kit/src/signing/index.tsx
@@ -2,9 +2,11 @@
 
 import type { Address, Hex } from 'viem'
 import { usePendingRequest } from './hooks/usePendingRequest.js'
+import { Erc20Approval } from './pages/Erc20Approval.js'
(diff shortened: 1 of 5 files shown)


packages/react-kit/src/signing/index.tsx
```diff
@@ -45,6 +47,21 @@ export function SignatureRequest() {
...
+              <Erc20Approval
+                contract={tx.to as Address}
+                spender={decoded.spender}
+                amount={decoded.amount}
+                confirm={confirm}
+                reject={reject}
+              />
+            )
+          }
+        }
```
OmarBasem, line 64: Should we return any error like `Failed to decode` if not decoded?

OmarBasem (approved):

packages/react-kit/src/signing/index.tsx
brtkx: The error is caught inside `decodeErc20Approval` and returns `null`, so we'll not have any errors here. In case of an error we will just show the generic screen.

PR #106, opened 2026-04-17
feat(react-kit): decode personal_sign messages in signing UI

brtkx:
closes FS-1964

### DIFF
--- a/packages/react-kit/src/signing/utils/personalSign.ts
+++ b/packages/react-kit/src/signing/utils/personalSign.ts
@@ -1,9 +1,9 @@
 import { type Hex, hexToString } from 'viem'
 
-export function decodePersonalSignMessage(data: Hex): string {
+export function decodePersonalSignMessage(data: Hex): string | null {
   try {
     return hexToString(data)
   } catch {
-    return data
+    return null
   }
 }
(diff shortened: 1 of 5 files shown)


packages/react-kit/src/signing/utils/personalSign.ts
```diff
@@ -0,0 +1,9 @@
+import { type Hex, hexToString } from 'viem'
+
+export function decodePersonalSignMessage(data: Hex): string {
+  try {
+    return hexToString(data)
+  } catch {
+    return data
+  }
+}
```
OmarBasem: Not sure if that's important, but function caller won't be able to distinguish if `hexToString` succeeded or is it just returning the same data (unless using an equality check)

OmarBasem (dismissed):

packages/react-kit/src/signing/utils/personalSign.ts
brtkx: @OmarBasem that makes sense to me, I've updated it [here](https://github.com/OffchainLabs/zerodev-wallet-sdk/pull/106/changes/343922136ec64485aa738ef06c44f5bb63f47d84) so that it returns null on failure and is consistent with all the other decoding

OmarBasem (approved):

PR #118, opened 2026-04-23
feat(react-kit): decode wallet_sendCalls batch transactions in signing UI

brtkx:
closes fs-1966

### DIFF
--- a/packages/react-kit/src/signing/pages/BatchCalls.tsx
+++ b/packages/react-kit/src/signing/pages/BatchCalls.tsx
@@ -1,0 +1,236 @@
+import {
+  type Address,
+  erc20Abi,
+  formatEther,
+  formatUnits,
+  type Hex,
+  maxUint256,
+} from 'viem'
+import { useReadContract } from 'wagmi'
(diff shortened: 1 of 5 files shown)


packages/react-kit/src/signing/pages/BatchCalls.tsx
```diff
@@ -0,0 +1,236 @@
...
+      </p>
+      <div className="mt-1 text-sm text-gray-500">
+        <span className="font-medium">To: </span>
+        <span className="font-mono break-all">{to}</span>
+      </div>
+    </div>
+  )
+}
+
+function Erc20ApprovalItem({
```
OmarBasem, line 79: Are these items meant to be components in different files?

brtkx: I'm not sure yet. Depends on how they are going to be presented in the designs. I think they may be just separate expandable cards - for now it makes sense keeping them here.

OmarBasem (dismissed):

OmarBasem (approved):

PR #123, opened 2026-04-24
feat(react-kit): trigger auth flow on connect instead of on init

brtkx:
closes FS-2089

Now auth flow will only show when `connect` is called, instead of displaying the screen right away.

### DIFF
--- a/apps/react-example/src/App.tsx
+++ b/apps/react-example/src/App.tsx
@@ -7,6 +7,7 @@ import {
 import { encodeFunctionData, erc20Abi, parseEther } from 'viem'
 import {
   useAccount,
+  useConnect,
   useDisconnect,
   useSendCalls,
   useSendTransaction,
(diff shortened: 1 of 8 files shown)


apps/react-example/src/App.tsx
```diff
@@ -193,10 +194,25 @@ function WalletPanel() {
...
+          <button
+            type="button"
+            onClick={() => connect({ connector: connectors[0] })}
+            className="mb-4 w-full rounded-lg bg-blue-600 px-4 py-2 text-white font-medium hover:bg-blue-700 cursor-pointer"
+          >
+            Connect
+          </button>
+          <AuthFlow />
+        </>
+      ) : (
```
OmarBasem, line 213: When the AuthFlow is displayed, shouldn't we hide the `connect` button?

brtkx: The app is just for our testing, doesn't really matter much here

OmarBasem (approved):

PR #124, opened 2026-04-24
feat(expo-example): add Passkey auth and test on Android

andrascodes:

* **#168**
* **#142**
* **#141**
* **#131**
* **#80**
* ➡️ **#124**
* `main`
</details></td></tr></table>

Closes FS-1928, FS-1999

The main purpose of the PR is setting up the Android build in a way that we can accomplish a PasskeyAuth on Android Emulator. Since the PasskeyStamper is already committed to the `expo-example` repo I just added a component to the app to trigger the Passkey register/login (using the hook) - no other changes were needed to make  it work.

Changes explained:
- `assetlinks.json`: a static json file deployed at `zerodev-expo-example.vercel.app`, this connects the domain to the app for App Links and Passkeys both
- `debug.keystore`: the app build needs to be signed with a key and the resulting fingerprint is added to the `assetlinks.json` file this makes the connection between the app and the domain - this keystore is committed so every contributor can use it to sign the builds and test Passkeys and App Links
- `withDebugKeystore.js`: expo plugin applied so the Android prebuilds use the `debug.keystore` instead of the one on the contributor's machine
- Separated the `wagmi.config` files for platform specific ones
- `eas.json`: setting up the EAS CLI for creating a local APK build that can be installed on a physical device or emulator, `credentials.json` is what this is looking for to get the signing credentials for the build
    - So there's the Expo Development build that you get by running `pnpm android` and a local APK build you get by running `pnpm dlx eas-cli build --platform android --local --profile preview`. Both will use the committed `debug.keystore` for signing so Passkeys and App Links will work on both (both looks at the `zerodev-expo-example.vercel.app` domain for verification).


apps/expo-example/plugins/withDebugKeystore.js
```diff
@@ -0,0 +1,24 @@
...
+ * Applied on every prebuild, so `pnpm android`, `npx expo run:android`,
+ * and `eas build` all produce APKs signed with the same cert.
+ */
+module.exports = function withDebugKeystore(config) {
+  return withAppBuildGradle(config, (config) => {
+    config.modResults.contents = config.modResults.contents.replace(
+      "storeFile file('debug.keystore')",
+      "storeFile file('../../credentials/debug.keystore')",
+    )
+    return config
```
SahilVasava, line 29: If Expo ever changes the prebuild template (different quote style, extra config block, etc.), this `replace` becomes a no-op and signing silently falls back to `~/.android/debug.keystore` per machine — passkeys then break with no useful error.

Worth asserting the replacement actually happened:

```js
const before = config.modResults.contents
config.modResults.contents = before.replace(
  "storeFile file('debug.keystore')",
  "storeFile file('../../credentials/debug.keystore')",
)
if (config.modResults.contents === before) {
  throw new Error("withDebugKeystore: storeFile pattern not found in build.gradle — Expo prebuild template may have changed")
}
```

Fails loud at prebuild time instead of silently breaking passkeys at runtime.

SahilVasava (commented):
Solid. One nit on the prebuild plugin to fail loud instead of silent.

SahilVasava (approved):
LGTM — the fail-loud check is now in `withDebugKeystore.js`: throws if the expected `storeFile file('debug.keystore')` line isn't found, so a template change won't silently degrade signing.

PR #129, opened 2026-04-27
feat(react-kit): migrate `sendTransaction` page UI

OmarBasem:
Closes [FS-1888](https://linear.app/offchain-labs/issue/FS-1888/ui-variant-for-token-standard-transfer)

### Summary

This PR implements UI for transfer send (ETH, and ERC20)

### DIFF
--- a/packages/react-kit/src/signing/index.tsx
+++ b/packages/react-kit/src/signing/index.tsx
@@ -1,5 +1,7 @@
 import type { Address, Hex } from 'viem'
 import { ScreenWrapper } from '../shared/components/ScreenWrapper'
+import type { GasFee, GasTier } from './components/TxGasFees/index.js'
+import type { Dapp } from './components/TxInformation/index.js'
 import { usePendingRequest } from './hooks/usePendingRequest.js'
 import { BatchCalls } from './pages/BatchCalls.js'
 import { CollectionApproval } from './pages/CollectionApproval.js'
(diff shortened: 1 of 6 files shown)


packages/react-kit/src/signing/index.tsx
```diff
@@ -17,12 +19,40 @@ import { decodeErc20Approval, isErc20Approval } from './utils/erc20Approval.js'
...
+  dapp: Dapp
+  selectedGasTier: GasTier
+  gasFees: GasFee[]
+  slippage?: number
+  tokenSubtitle: string
+  tokenImageSource: string
+  recipientImageSource: string
+}
+
+export function SignatureRequest({
```
brtkx: We can't do this. The dev would have to provide all these props when mounting `SignatureRequest` - but our selling point is that they only need to mount it without extra props.

We should instead handle those inside each signing page. E.g. in `EthTransfer.tsx` we'd do something like:
```typescript
<InfoCard
  title={`${formattedAmount} ETH`}
  subtitle="$0.01"
  imageSource="hardcode token image for now"
```

We can just hardcode those details, but they can't be passed as props to the top level component.

packages/react-kit/src/signing/pages/EthTransfer.tsx
```diff
@@ -1,27 +1,74 @@
...
-            <span className="font-mono break-all">{to}</span>
-          </div>
+      <div className="flex flex-col gap-2 pt-4">
+        <div className="flex flex-col items-center justify-center gap-2 pb-2">
+          <Text className="text-h2">Send Token</Text>
+          <Text className="text-center">
+            You are about to send {formattedAmount} ETH to {to}.
+          </Text>
+        </div>
+        <TxInformation dapp={dapp} />
```
brtkx: We can remove `dapp` from all components. We don't need to show it because the wallet will always live in the app's domain, not an iframe.

brtkx (approved):

PR #130, opened 2026-04-28
feat(react-kit): SignatureRequest usage modes

brtkx:
closes FS-1891

`<SignatureRequest />` can now be used in different modes:
1. Simply mounting and all signing displays automatically (like so far):
```typescript
<SignatureRequest />
```
2. Using props to create a custom UI:
```typescript
<SignatureRequest>
  {({ pendingRequest, confirm, reject }) => (
    <CustomConfirmUI
      pendingRequest={pendingRequest}
      confirm={confirm}
      reject={reject}
    />
  )}
</SignatureRequest>
```
3. Passing request props - keeps the same UI but isn't tied to automatic request tracking. Dev can display whatever request they want here.
```typescript
<SignatureRequest
  request={controlledRequest}
  onConfirm={handleControlledConfirm}
  onReject={() => setControlledRequest(null)}
/>
```

All modes testable in the example app. Next I can do the same for authentication.

### DIFF
--- a/apps/react-example/src/App.tsx
+++ b/apps/react-example/src/App.tsx
@@ -1,10 +1,12 @@
 import {
   AuthFlow,
   Button,
+  type PendingRequest,
+  type Request,
   SignatureRequest,
   usePendingRequest,
 } from '@zerodev/wallet-react-kit'
-
+import { useId, useState } from 'react'
 import { encodeFunctionData, erc20Abi, parseEther } from 'viem'
 import {
   useAccount,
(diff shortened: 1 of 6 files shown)


apps/react-example/src/App.tsx
```diff
@@ -31,6 +74,31 @@ function WalletPanel() {
...
+        {
+          from: address!,
+          to: address!,
+          value: `0x${parseEther('0.01').toString(16)}`,
+        },
+      ],
+    })
+  }
+
+  function handleControlledConfirm() {
```
OmarBasem, line 96: Currently the default and controlled looks similar, maybe show a feedback that shows the difference like an alert for the controlled confirm saying something like "handling a controlled request"

brtkx: As long as we can test things it's fine imo. This will all be removed and added to zerodev example app anyway

OmarBasem (approved):

PR #131, opened 2026-04-28
feat(expo-example): migrate to expo-router

andrascodes:

* **#168**
* **#142**
* **#141**
* ➡️ **#131**
* `main`
</details></td></tr></table>

Closes FS-2099

Decided to migrate the `expo-example` app to Expo Router to make the app resemble a production-grade app a bit more closely. Also, because the Magic Link callback I'm implementing in a subsequent PR works better using Expo Router.
This also caused the appearance of some bugs:
1) Wagmi automatic reconnection triggers a React `set-state-in-render` warning, which I handled with opting out of `reconnectOnMount` and triggering the `reconnect()` manually in `_layout`
2) The `/oauth-callback` was processed by Expo-Router before (or in parallel) with the Native OAuth flow. I fixed it by just adding an explicit route for it and rendering an `ActivityIndicator` meanwhile the `Stack.Protected` usage in `_layout` takes care of the redirection to the authenticated parts of the app. This can serve as an example of solving this issue in prod.


apps/expo-example/src/app/_layout.tsx
```diff
@@ -0,0 +1,77 @@
+import 'react-native-get-random-values'
```
andrascodes, line 1: This is basically the only file that needs to be reviewed in this PR. The rest is just moving the files around (and fixing file references) or following the Expo docs on setting up Expo Router.
This file implements the Protected routes (authenticated routes) and fixes a bug around using Wagmi on React Native which causes an error in development (`set-state-in-render` - manual reconnection solves it)
The SplashScreen should show until we can determine the authenticated status. This doesn't work on Android preview because of Expo. It only works in a production build.

OmarBasem (approved):

PR #144, opened 2026-04-30
feat(react-kit): replace hardcoded gas fees with UserOp estimate

brtkx:
closes FS-2113

### DIFF
--- a/packages/react-kit/src/signing/pages/EthTransfer.tsx
+++ b/packages/react-kit/src/signing/pages/EthTransfer.tsx
@@ -2,21 +2,16 @@ import { type Address, formatEther, type Hex } from 'viem'
 
 import { Text } from '../../shared/components/Text'
 import { ArrowCardPair } from '../components/ArrowCardPair'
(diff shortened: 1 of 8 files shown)


packages/react-kit/src/signing/pages/EthTransfer.tsx
```diff
@@ -55,10 +62,16 @@ export function EthTransfer({ to, value, confirm, reject }: EthTransferProps) {
               />
             }
           />
-          <TxGasFees
-            selectedGasTier={SELECTED_GAS_TIER}
-            gasFees={GAS_FEES}
-            slippage={SLIPPAGE}
+          <DataRow
```
brtkx, line 65: The reason `TxGasFees` is removed is because we cannot show gas tiers right now - in Doorway that was an oversight as well.

Currently sending any transaction goes through user op meaning we cannot estimate gas with wagmi only - we need to use ZD method for estimation which is a single tier result.

OmarBasem (approved):

PR #150, opened 2026-05-05
feat(react-kit): add `magicLink` and set auth method

OmarBasem:
Closes [FS-2132](https://linear.app/offchain-labs/issue/FS-2132/separate-otp-and-magic-link-flows)

### Summary

- This PR adds magic link flow
- Separates OTP and magic link flows. It adds `emailAuthMethod` as an auth config option, where the default is magic link

### DIFF
--- a/packages/react-kit/src/auth/authStoreSlice.ts
+++ b/packages/react-kit/src/auth/authStoreSlice.ts
@@ -1,6 +1,47 @@
 import type { StateCreator } from 'zustand'
 import type { AuthConfig, AuthMethod, AuthStep } from './types'
 
+const OTP_SESSION_STORAGE_KEY = 'zerodev:auth:otpSession'
+
+type StoredOtpSession = {
+  otpId: string
+  otpEncryptionTargetBundle: string
+}
+
+function readStoredOtpSession(): StoredOtpSession | null {
+  if (typeof window === 'undefined') return null
+  try {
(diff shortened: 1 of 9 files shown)


packages/react-kit/src/auth/authStoreSlice.ts
```diff
@@ -1,6 +1,47 @@
 import type { StateCreator } from 'zustand'
 import type { AuthConfig, AuthMethod, AuthStep } from './types'
 
+const OTP_SESSION_STORAGE_KEY = 'zerodev:auth:otpSession'
```
OmarBasem, line 4: Store otp session data in local storage for magic link flow as verification opens in a new tab

packages/react-kit/src/auth/pages/SignUp.tsx
```diff
@@ -22,7 +26,11 @@ export function SignUp() {
   const { goToStep, setEmail, setOtpSession, config } = useAuth()
   const [agreedToTerms, setAgreedToTerms] = useState(false)
   const [emailInput, setEmailInput] = useState('')
-  const { mutateAsync: sendOtp, isPending: isEmailLoading } = useSendOTP()
+  const useOtp = config?.emailAuthMethod === 'otp'
```
brtkx: nit: `shouldUseOtp` just so it doesn't follow the hook naming convention

brtkx (dismissed):

brtkx (approved):

PR #154, opened 2026-05-06
feat(react-kit): expand docs demo and fix embedded rendering

brtkx:
closes fs-2142

  - Expand the docs `<SignatureRequest />` live demo to mirror `apps/react-example`'s button set: ETH/ERC-20 send, ERC-20 approve, collection approval, sign message, sign typed data, batch calls.
  - Fix `ScreenWrapper` rendering in non-fullscreen embeds (the kind of host layout the docs use, but also any consumer dropping the kit into a panel rather than a phone-frame parent):
    - `MultiRadialBackground` `-z-20` → `z-0`. The negative z-index escaped `ScreenWrapper`'s non-stacking-context parent and pushed the gradient SVG behind unrelated page chrome.
    - Added `max-w-[500px] rounded-[30px]` to `ScreenWrapper` so the panel keeps phone-frame proportions in wide hosts.
    - `<AppLogo>` on auth pages: `absolute bottom-6` → flow positioning, since absolute positioning ignored the new max-width and overlapped content.
  - Hide the "Choose a wallet instead" entry on `SignUp` (commented, not deleted) — the `wallet-selection` step still renders a placeholder, so the entry was a dead end.
  - Fix disconnect button text in the docs (was inheriting white from vocs's page color).

### DIFF
--- a/packages/react-kit/src/auth/pages/EmailVerification.tsx
+++ b/packages/react-kit/src/auth/pages/EmailVerification.tsx
@@ -73,7 +73,7 @@ export function EmailVerification() {
             </Text>
           </div>
 
-          <AppLogo className="absolute self-center bottom-6" />
+          <AppLogo className="self-center pb-6" />
         </div>
       )}
     </ScreenWrapper>
--- a/packages/react-kit/src/auth/pages/EmailVerification.tsx
+++ b/packages/react-kit/src/auth/pages/EmailVerification.tsx
@@ -46,34 +46,36 @@ export function EmailVerification() {
       {({ paddingTop }) => (
         <div
           style={{ paddingTop: `${paddingTop}px` }}
-          className="flex flex-1 flex-col gap-8 justify-center h-full"
+          className="flex flex-1 flex-col h-full"
         >
-          <StatusView
-            imageName="send"
-            title={'Check your email!\n An Email is On Its Way'}
-          >
-            We've sent a magic link to{' '}
-            <Text className="text-solarOrange inline">{email}</Text>
-            {'\n'}Please open the email and click the link to log in.
-          </StatusView>
+          <div className="flex-1 flex flex-col gap-8 justify-center">
+            <StatusView
+              imageName="send"
+              title={'Check your email!\n An Email is On Its Way'}
+            >
+              We've sent a magic link to{' '}
+              <Text className="text-solarOrange inline">{email}</Text>
+              {'\n'}Please open the email and click the link to log in.
+            </StatusView>
 
-          <div className="flex flex-col gap-1">
-            <Text className="text-center">
-              Did not get an email?{' '}
-              <button
-                type="button"
-                disabled={!canResend}
-                onClick={handleResend}
-                className="cursor-pointer underline disabled:opacity-50 disabled:cursor-not-allowed"
-              >
-                {canResend
-                  ? 'Resend'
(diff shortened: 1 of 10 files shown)


packages/react-kit/src/auth/pages/EmailVerification.tsx
```diff
@@ -73,7 +73,7 @@ export function EmailVerification() {
             </Text>
           </div>
 
-          <AppLogo className="absolute self-center bottom-6" />
+          <AppLogo className="self-center pb-6" />
```
brtkx: I don't know if absolute is intended here @OmarBasem but it breaks some layouts. Prefer to change it, lmk

packages/react-kit/src/auth/pages/SignUp.tsx
```diff
@@ -167,14 +163,14 @@ export function SignUp() {
                 </Input>
               </div>
               <OrView />
-              <ListItem
+              {/* <ListItem
```
brtkx, line 166: Disabling this for now because we only support ZD wallet in alpha

packages/react-kit/src/shared/components/ScreenWrapper/index.tsx
```diff
@@ -26,7 +26,7 @@ export function ScreenWrapper({
   return (
     <div
       className={cn(
-        'flex-1 flex flex-col relative overflow-hidden h-full',
+        'flex-1 flex flex-col relative overflow-hidden h-full max-w-[500px] rounded-[30px]',
```
brtkx: Makes it prettier

packages/react-kit/src/shared/components/ScreenWrapper/MultiRadialBackground.tsx
```diff
@@ -65,7 +65,7 @@ export function MultiRadialBackground() {
   ]
 
   return (
-    <div className="absolute inset-0 -z-20 rounded-[36px] pointer-events-none">
+    <div className="absolute inset-0 z-0 rounded-[36px] pointer-events-none">
```
brtkx, line 68: @OmarBasem idk if -z-20 is needed - I don't think so. It broke the background in docs, which I think other devs could face as well. z-index 0 works for react example too.

packages/react-kit/src/auth/pages/EmailVerification.tsx
OmarBasem: Yes `absolute` is needed to keep it at the bottom of the screen, what layouts does it break?

packages/react-kit/src/shared/components/ScreenWrapper/index.tsx
OmarBasem: Make it `rounded-[34px]` to look more aligned from the corners

packages/react-kit/src/auth/pages/EmailVerification.tsx
brtkx: In react example we wrap `AuthFlow` with a div that has `h-[800px]` on it. If we don't do this, then `AppLogo` overlaps content. We can still keep `AppLogo` at the bottom of the page while having it non absolute. But we shouldn't have devs needing to wrap our components with fixed height to fix broken UI.

packages/react-kit/src/shared/components/ScreenWrapper/index.tsx
brtkx: done

OmarBasem (approved):

PR #166, opened 2026-05-11
feat(react-kit): loading skeletons across signing pages

brtkx:
## Summary

  - Signing pages previously bailed to bare text (`"Loading token details..."`, `"Estimating..."`) while async data resolved. Now they render a generic `<SigningPageSkeleton />` inside the signing chrome — Reject stays usable, Confirm is disabled, page chrome stays consistent. The skeleton's shape (two card placeholders + a single fee-row skeleton) matches the actual page surface so the swap doesn't snap to a different layout.
  - Gas-only refreshes get a narrow `<DataRowSkeleton />` swap on just the fee row, not the whole page. Gas refetches periodically — flashing the full page skeleton on each refresh would be jarring, so the page-loading condition explicitly excludes gas.
  - `BatchCalls` and `GenericRequest` were displaying hardcoded fee strings (`"0.00008 ETH ($0.28)"`) regardless of network conditions — placeholder values that would mislead consumers. Both now call `useGasEstimate` against the real userop. The hook itself was refactored to take `calls: UseGasEstimateCall[]` (matching what the kernel client consumes) so batch and single-call pages share the same code path.
  - In `BatchCalls`, hoisted the per-item ERC-20 `useReadContract` calls out of the child components and into a single `useReadContracts` at the parent. The page now waits for all token metadata before rendering items, instead of each child decoding-and-flickering in sequence as its queries land.
  - Error fallback consistency: pages that previously bailed to bare `<Text>Failed to load token details</Text>` now wrap that in `<SigningLayout error=...>` so the user can still Reject — no more dead-end screens.

### DIFF
--- a/packages/react-kit/src/signing/components/SigningPageSkeleton/SigningPageSkeleton.stories.tsx
+++ b/packages/react-kit/src/signing/components/SigningPageSkeleton/SigningPageSkeleton.stories.tsx
@@ -1,0 +1,24 @@
+import type { Meta, StoryObj } from '@storybook/react-vite'
+
+import { SigningPageSkeleton } from '.'
+
+const meta = {
+  title: 'Signing/SigningPageSkeleton',
+  component: SigningPageSkeleton,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs'],
+  decorators: [
+    (Story) => (
+      <div style={{ width: 400 }}>
+        <Story />
+      </div>
+    ),
+  ],
+} satisfies Meta<typeof SigningPageSkeleton>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {}
--- a/packages/react-kit/src/signing/hooks/useGasEstimate.ts
+++ b/packages/react-kit/src/signing/hooks/useGasEstimate.ts
@@ -5,18 +5,22 @@ import type { ZeroDevWalletState } from '@zerodev/wallet-react'
 import type { Address, Hex } from 'viem'
 import { useChainId, useConfig } from 'wagmi'
 
-type UseGasEstimateParams = {
+export type UseGasEstimateCall = {
   to: Address
-  value?: Hex
-  data?: Hex
+  value?: Hex | undefined
+  data?: Hex | undefined
 }
 
-export function useGasEstimate({ to, value, data }: UseGasEstimateParams) {
+type UseGasEstimateParams = {
+  calls: UseGasEstimateCall[]
+}
+
+export function useGasEstimate({ calls }: UseGasEstimateParams) {
   const config = useConfig()
   const chainId = useChainId()
 
   return useQuery({
-    queryKey: ['userop-gas-estimate', chainId, to, value, data],
+    queryKey: ['userop-gas-estimate', chainId, calls],
     refetchInterval: 12_000,
     retry: false,
     queryFn: async (): Promise<bigint | null> => {
@@ -38,13 +42,11 @@ export function useGasEstimate({ to, value, data }: UseGasEstimateParams) {
       // misleading.
       const userOp = await kernelClient.prepareUserOperation({
         account,
-        calls: [
-          {
-            to,
-            value: value ? BigInt(value) : 0n,
-            data: data ?? '0x',
-          },
-        ],
+        calls: calls.map((c) => ({
+          to: c.to,
+          value: c.value ? BigInt(c.value) : 0n,
+          data: c.data ?? '0x',
+        })),
       })
       const totalGas =
         userOp.preVerificationGas +
--- a/packages/react-kit/src/signing/pages/BatchCalls.tsx
+++ b/packages/react-kit/src/signing/pages/BatchCalls.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from 'react'
 import {
   type Address,
   erc20Abi,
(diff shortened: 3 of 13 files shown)


packages/react-kit/src/signing/components/SigningPageSkeleton/SigningPageSkeleton.stories.tsx
brtkx, line 1: One shared loading component for all signing screens for now.

packages/react-kit/src/signing/hooks/useGasEstimate.ts
```diff
@@ -5,18 +5,22 @@ import type { ZeroDevWalletState } from '@zerodev/wallet-react'
...
+  value?: Hex | undefined
+  data?: Hex | undefined
+}
+
+type UseGasEstimateParams = {
+  calls: UseGasEstimateCall[]
 }
 
-export function useGasEstimate({ to, value, data }: UseGasEstimateParams) {
+export function useGasEstimate({ calls }: UseGasEstimateParams) {
```
brtkx, line 18: This change allows batch calls to also estimate gas

packages/react-kit/src/signing/pages/BatchCalls.tsx
brtkx, line 1: These changes allow the loading component to be displayed for batch calls

packages/react-kit/src/signing/pages/BatchCalls.tsx
```diff
@@ -242,12 +318,23 @@ export function BatchCalls({ calls, confirm, reject }: BatchCallsProps) {
...
                 call={call}
                 index={i + 1}
+                tokenInfoMap={tokenInfoMap}
               />
             ))}
           </div>
         </DetailsContainer>
         <DetailsContainer title="Estimated Gas Fee" iconName="lightingFill">
-          <DataRow label="Fee" value={GAS_FEE} iconName="gasStation" />
+          {gasError ? (
```
brtkx, line 327: `gasError` will be removed in the future PR because we will show an info box with a wagmi error instead. Applies to all signing components

packages/react-kit/src/signing/pages/Erc20Approval.tsx
```diff
@@ -49,18 +50,29 @@ export function Erc20Approval({
...
+    )
   }
 
   if (!decimals || !symbol) {
-    return <Text>Failed to load token details.</Text>
+    return (
+      <SigningLayout
+        onConfirm={confirm}
+        onReject={reject}
+        error={new Error('Failed to load token details')}
```
brtkx, line 71: Will use wagmi error here instead too in the future PR

packages/react-kit/src/signing/pages/BatchCalls.tsx
```diff
@@ -225,9 +226,84 @@ function CallItem({ call, index }: { call: BatchCall; index: number }) {
...
+
+  if (tokensLoading) {
+    return (
+      <SigningLayout onConfirm={confirm} onReject={reject} disabled>
+        <SigningPageSkeleton />
+      </SigningLayout>
+    )
+  }
+
+  const confirmDisabled = gasFetching || gasEstimate == null
```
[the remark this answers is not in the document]
brtkx: nah we don't want to allow signing when gas is re-estimated

packages/react-kit/src/signing/pages/CollectionApproval.tsx
```diff
@@ -106,17 +110,17 @@ export function CollectionApproval({
...
-            iconName="gasStation"
-          />
+          {gasError ? (
+            <DataRow label="Network fee" value="Error" iconName="gasStation" />
+          ) : gasEstimate != null ? (
+            <DataRow
+              label="Network fee"
+              value={formatGasFee(gasEstimate)}
+              iconName="gasStation"
+            />
```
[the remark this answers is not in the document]
brtkx: `gasError` will be removed anyway in the next PR and replaced with a info box error

packages/react-kit/src/signing/pages/BatchCalls.tsx
```diff
@@ -183,12 +182,13 @@ function CallItem({ call, index }: { call: BatchCall; index: number }) {
...
     const decoded = decodeErc20Transfer(tx)
     if (decoded) {
+      const contract = call.to as Address
       return (
         <Erc20TransferItem
-          contract={call.to as Address}
           to={decoded.to}
           amount={decoded.amount}
           index={index}
+          tokenInfo={tokenInfoMap.get(contract) ?? {}}
```
OmarBasem, line 191: Could it be a problem if `tokenInfo` is `{}` ?

OmarBasem (commented):
Code looks good overall, but the Gas estimate skeleton UI I think is not very visible, maybe something like "Estimating gas..." with a loading indicator would look better

packages/react-kit/src/signing/pages/BatchCalls.tsx
brtkx: Not really a problem. If we get `{}` here then the UI will show number of tokens (not amount) if decimals are missing, or `ERC-20` text if symbol is missing. But it's unlikely fallback - we should be getting data

brtkx: We can improve those default values later, we have several places where it could be better

OmarBasem (approved):

PR #168, opened 2026-05-11
feat(expo-example): add export wallet on mobile

andrascodes:
Closes FS-1979

For the export wallet on mobile, I decided to go with the same approach as the `iframe-stamper` on Web. This is the `ExportWebView` file. It requests the bundle using the `wallet-core` export functions and passes the bundle to the `export.turnkey.com` website via an iframe. The iframe on mobile is rendered in a WebView. The `ExportWebView` component injects some JS into the WebView to enable passing the bundle to the iframe itself via a `MessageChannel`.
Other than this the rest of the files are to demo certain further security measures that app developers can take: Biometrics before revealing the key, blocking screen capture both for screenshots and for not showing the keys when switching between apps and protection against supply chain attacks (restricting package versions and network security config)


OmarBasem (approved):
Tested on physical devices on iOS and Android, most auth flows are working except for dynamic links on both platforms, in addition to passkeys and universal links on iOS which needs extra setup (discussed on Slack, yet to be setup and tested out)

Export wallet looks good

Transactions looks good

A small comment on the email input keyboard, should not hide the email input

PR #174, opened 2026-05-13
fix(react-kit): highlight terms and check email

OmarBasem:
Closes [FS-2179](https://linear.app/offchain-labs/issue/FS-2179/fix-tandc-checkbox-so-its-clear-to-user-they-need-to-accept-it)

### Summary

This PR fixes the following:

- `handleEmailSubmit` button will be always rendered (after starting to type) but will stay disabled until a valid email is entered, and terms are agreed to (if set in config)
- If the user tried to press the `handleEmailSubmit` button after entering a valid email and before agreeing to terms, the agree to terms box with be highlighted with a red border

### DIFF
--- a/packages/react-kit/src/auth/pages/SignUp.tsx
+++ b/packages/react-kit/src/auth/pages/SignUp.tsx
@@ -14,6 +14,7 @@ import { OrView } from '../../shared/components/OrView'
 import { ScreenWrapper } from '../../shared/components/ScreenWrapper'
 import { SignUpFooter } from '../../shared/components/SignUpFooter'
 import { Text } from '../../shared/components/Text'
+import { isValidEmailAddress } from '../../shared/utils/common'
 import { useAuth } from '../hooks/useAuth'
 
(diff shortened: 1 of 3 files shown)


packages/react-kit/src/auth/pages/SignUp.tsx
```diff
@@ -130,43 +132,35 @@ export function SignUp() {
...
-      goToStep('otp-input')
-    } catch (err) {
-      setError(
-        err instanceof Error ? err.message : 'Failed to send verification code',
-      )
+  const handleEmailSubmit = async (forceMethod?: 'otp' | 'magicLink') => {
+    if (!emailInput || anyPending) return
+    if (!isValidEmailAddress(emailInput)) return
+    if (!canContinue) {
+      setHighlightAgreement(true)
```
brtkx, line 150: Let's also add it to all the other login methods (except for injected wallets)

brtkx (approved):

PR #176, opened 2026-05-13
fix(react-kit): logout empty screen

OmarBasem:
Closes [FS-2182](https://linear.app/offchain-labs/issue/FS-2182/fix-empty-screen-after-disconnectlogout)

### Summary

This PR fixes the following:

- Having an empty screen after disconnect/logout, should render sign-up page

### DIFF
--- a/packages/react-kit/src/connector.ts
+++ b/packages/react-kit/src/connector.ts
@@ -85,10 +85,13 @@ export function zeroDevWallet(
 
       async disconnect() {
         await connector.disconnect?.()
-        // Reset auth state on disconnect
+        // Reset auth state on disconnect and surface the sign-up page so the
+        // user can immediately re-authenticate instead of seeing an empty
+        // AuthFlow (the 'initializing' step renders null).
         if (params.config?.auth) {
           store.getState().auth.reset()
           store.getState().auth.initialize(params.config.auth)
+          store.getState().auth.goToStep('sign-up')
         }
       },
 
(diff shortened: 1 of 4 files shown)


packages/react-kit/src/connector.ts
```diff
@@ -85,10 +85,13 @@ export function zeroDevWallet(
...
       async disconnect() {
         await connector.disconnect?.()
-        // Reset auth state on disconnect
+        // Reset auth state on disconnect and surface the sign-up page so the
+        // user can immediately re-authenticate instead of seeing an empty
+        // AuthFlow (the 'initializing' step renders null).
         if (params.config?.auth) {
           store.getState().auth.reset()
           store.getState().auth.initialize(params.config.auth)
+          store.getState().auth.goToStep('sign-up')
```
OmarBasem, line 94: I was thinking if we even need the `initializing` step, we can consider removing it, I don't think the `sign-up` page ui depends on the anything from the initialize step

brtkx: Doesn't `initializing` step simply mean "don't render AuthFlow"? I think we still need this behavior? If so then it's just a case of naming this step something more accurate

brtkx: Unless we allow the step to be `null` and then we render `null` - I kinda like this solution

OmarBasem: I will consider refactoring in another PR

brtkx (approved):

PR #177, opened 2026-05-13
feat(react-kit): surface gas estimate errors in AlertView

brtkx:
closes FS-2175

- Route gas estimation errors through `SigningLayout`'s existing AlertView surface so the confirm button gets disabled.
- Decode embedded `Error(string)` revert hex (selector `0x08c379a0`) before display. ERC-4337 errors come from the bundler as `"...UserOperation reverted during simulation with reason: 0x..."` and viem doesn't decode the inner string for us.
- Hide the gas DataRow entirely on error — the AlertView covers the context, the row was redundant.
- `break-all` on AlertView description so long unbreakable strings (hex, addresses) wrap instead of overflowing the box horizontally.

### DIFF
--- a/packages/react-kit/src/signing/components/SigningLayout/index.tsx
+++ b/packages/react-kit/src/signing/components/SigningLayout/index.tsx
@@ -1,51 +1,8 @@
 import type { ReactNode } from 'react'
-import { BaseError, decodeErrorResult, type Hex } from 'viem'
 
 import { AlertView } from '../../../shared/components/AlertView'
+import { getTxErrorInfo } from '../../utils/errors'
 import { SigningActions } from '../SigningActions'
-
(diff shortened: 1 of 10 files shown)


packages/react-kit/src/signing/components/SigningLayout/index.tsx
```diff
@@ -9,13 +9,30 @@ interface TxErrorInfo {
   description: string
 }
 
+// Replace any embedded `Error(string)` revert hex (selector 0x08c379a0)
+// with its decoded string. Bundler-wrapped errors surface the raw hex.
+function decodeErrorHex(message: string): string {
```
OmarBasem: Can be moved to utils

OmarBasem (approved):

PR #181, opened 2026-05-14
feat(react-kit): TopNav

OmarBasem:
Closes [FS-2180](https://linear.app/offchain-labs/issue/FS-2180/allow-going-back-between-auth-steps-and-allow-closing-the-auth-window)

### Summary

This PR does the following updates:

- Adds a `TopNav` component, that includes an "X" button to close the AuthFlow, a back button when navigating to any page after sign-up page, and can optionally have a title
- Simplifies `ScreenWrapper` usage in AuthFlow, now it wraps the `index` file instead of each individual page, will do the same for `SignatureRequest` in a follow-up PR


packages/react-kit/src/shared/components/ScreenWrapper/index.tsx
```diff
@@ -33,6 +35,7 @@ export function ScreenWrapper({
     >
       <MultiRadialBackground />
       <div className="flex-1 bg-offWhite/85 m-1.5 px-4 overflow-hidden rounded-[30px] relative">
+        {topNav}
         {children({ paddingTop, paddingBottom })}
```
brtkx: Wouldn't this work so we don't have to pass padding to children?
```typescript
{topNav}
<div style={{ paddingTop: `${paddingTop}px` }}>
  {children}
</div>
```

apps/zerodev-signer-demo/src/app/wagmi-config.ts
```diff
@@ -26,8 +26,10 @@ export const config = createConfig({
       config: {
         auth: {
           magicLinkBaseUrl: `${process.env.NEXT_PUBLIC_APP_URL ?? 'http://localhost:3000'}/verify`,
-          enabledMethods: ['email', 'google', 'passkey'],
-          emailAuthMethod: 'otp'
+          enabledMethods: ['email', 'google', 'passkey', 'injected-wallet'],
+          emailAuthMethod: 'otp',
+          termsAndConditionsUrl: 'https://zerodev.app/terms-of-service',
+          privacyPolicyUrl: 'https://zerodev.app/privacy-policy',
```
OmarBasem, line 32: Should we have these urls here for demoing purposes?

brtkx: We can

brtkx (approved):

PR #190, opened 2026-05-15
feat(demo): toggle SignatureRequest mount

brtkx:
Add a "Show transaction review" checkbox to the dashboard wallet card. Toggling it mounts/unmounts `<SignatureRequest />`


OmarBasem (commented):
Why are we adding this toggle, don't we want the signature request to be handled by react-kit?

brtkx:
> Why are we adding this toggle, don't we want the signature request to be handled by react-kit?

@OmarBasem yes but for the zerodev app, product wants to have a toggle so we can showcase both cases

OmarBasem (approved):

PR #192, opened 2026-05-15
chore(demo): drop injected-wallet from enabledMethods

brtkx:


apps/zerodev-signer-demo/src/app/wagmi-config.ts
```diff
@@ -26,7 +26,7 @@ export const config = createConfig({
       config: {
         auth: {
           magicLinkBaseUrl: `${process.env.NEXT_PUBLIC_APP_URL ?? 'http://localhost:3000'}/verify`,
-          enabledMethods: ['email', 'google', 'passkey', 'injected-wallet'],
+          enabledMethods: ['email', 'google', 'passkey'],
```
OmarBasem, line 29: I remember Srinjoy said we want to support external wallets for alpha?

brtkx: yes but not in the demo app I think, I will confirm again

PR #202, opened 2026-05-19
feat(expo-example): ios passkeys and universal links

OmarBasem:
Closes [FS-2198](https://linear.app/offchain-labs/issue/FS-2198/add-required-setup-for-passkeys-and-universal-links-on-ios)

### Summary

This PR adds the following:

- Required setup for passkeys iOS
- Required setup for universal links on iOS
- Defaults to using app/universal links
- Documentation for iOS passkeys and universal links

Tested authentication using passkeys and universal links on an iOS physical device.


apps/expo-example/src/config/auth.ts
```diff
@@ -4,10 +4,8 @@ import * as Linking from 'expo-linking'
 // (shared across contributors; see README "Running with Passkeys on Android").
 export const RP_ID = 'zerodev-expo-example.vercel.app'
 
-const shouldUseAppLinks = process.env.EXPO_PUBLIC_USE_APP_LINKS === 'true'
-
 export const buildRedirectUri = (path: string): string =>
-  shouldUseAppLinks ? `https://${RP_ID}/${path}` : Linking.createURL(path)
```
andrascodes, line 10: Could we default to App Links in the sense that we keep the option of using custom scheme as well? So Just flip the wording and meaning of the env var and if `EXPO_PUBLIC_USE_CUSTOM_SCHEME` is true then we use `Linking.createURL(path)`. But if it's not set it uses App Links by default.
I agree with your reasoning that the setup is already there for App Links and Custom scheme doesn't actually work for Magic Links nor is it recommended for OAuth. But this is not a prod app just an example and a test app, so I want to make sure we show to consumers that they can use OAuth with Custom Scheme as well if they want to. But we can make sure to recommend the usage of App Links on the docs and discourage using Custom Scheme in prod.

andrascodes (changes requested):
Small change request just to put back the Custom Scheme demo. But otherwise all good. Thanks so much for setting up and testing on iOS!! 🙏

andrascodes (approved):
Awesome!! Thanks!!

PR #206, opened 2026-05-19
fix(auth): null goBack when no history; strip ?code on retry/close

brtkx:


packages/react-kit/src/auth/hooks/useAuth.ts
```diff
@@ -37,7 +40,7 @@ export function useAuth() {
     enabledMethods,
     config: authConfig,
     goToStep: store.getState().auth.goToStep,
-    goBack: store.getState().auth.goBack,
+    onBack: stepHistory.length > 0 ? goBack : null,
```
brtkx: If can't go back, this will be `null` so we can use it in the UI

packages/react-kit/src/auth/pages/Verifying.tsx
```diff
@@ -19,60 +20,75 @@ export function Verifying() {
...
+  const { mutate: verifyMagicLink, isPending: isVerificationLoading } =
+    useVerifyMagicLink({
+      mutation: {
+        onSuccess: async () => {
+          clearOtpSession()
+          goToStep('authenticated')
+          config?.onSuccess?.()
+        },
+        onError: (err) => {
+          setError(err)
```
brtkx, line 36: `verificationError` doesn't expose all errors for some reason. For example, when `otpId` or `otpEncryptionTargetBundle` is missing, it will be stack on loading, but `onError` will give error.

packages/react-kit/src/auth/pages/Verifying.tsx
```diff
@@ -19,60 +20,75 @@ export function Verifying() {
...
 
   useEffect(() => {
-    if (hasVerifiedRef.current || !otpId || !otpEncryptionTargetBundle || !code)
-      return
+    if (hasVerifiedRef.current || !code) return
 
     hasVerifiedRef.current = true
-    verifyMagicLink({ otpId, code, otpEncryptionTargetBundle })
+    verifyMagicLink({
+      otpId: otpId ?? '',
```
brtkx, line 47: Letting empty `otpId` and `otpEncryptionTargetBundle` because when users land on an invalid code we want the function to run and give error. Otherwise we end up with an empty verification screen

packages/react-kit/src/auth/authStoreSlice.ts
```diff
@@ -110,15 +110,19 @@ export const createAuthStoreSlice: StateCreator<
         auth: {
           ...state.auth,
           step,
-          stepHistory: [...state.auth.stepHistory, state.auth.step],
+          stepHistory:
```
brtkx, line 113: We don't add `initializing` the history and we don't default to `sign-up`. `goBack` now can be `null` if nothing is in history so we don't need to default

OmarBasem (approved):

PR #215, opened 2026-05-20
fix(react-kit): error on re-auth

OmarBasem:
### Summary

This PR fixes an error where if the user tried to press on the magic link again after already being authenticated then get redirected to the verify page again, then an error briefly appears, before being redirected to the dashboard. Now if the user is already authenticated it will go to the dashboard directly.

### DIFF
--- a/packages/react-kit/src/auth/pages/Verifying.tsx
+++ b/packages/react-kit/src/auth/pages/Verifying.tsx
@@ -42,13 +42,25 @@ export function Verifying() {
   useEffect(() => {
     if (hasVerifiedRef.current || !code) return
 
+    // No active OTP session — most commonly the user tapped the same
+    // magic link again after a successful verify, which cleared the
+    // session. Skip the (doomed) mutation, clean the URL, and drop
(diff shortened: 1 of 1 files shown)


packages/react-kit/src/auth/pages/Verifying.tsx
```diff
@@ -42,13 +42,25 @@ export function Verifying() {
   useEffect(() => {
     if (hasVerifiedRef.current || !code) return
 
+    // No active OTP session — most commonly the user tapped the same
+    // magic link again after a successful verify, which cleared the
+    // session. Skip the (doomed) mutation, clean the URL, and drop
+    // back to step=null so the host app's own routing handles the
+    // already-authenticated user without flashing an error.
+    if (!otpId || !otpEncryptionTargetBundle) {
+      hasVerifiedRef.current = true
```
brtkx: nit: we can move the same line from L57 above this if statement and remove this one

brtkx (approved):

PR #227, opened 2026-05-21
feat(core): send PoP signature on OAuth completion

SahilVasava:
## Summary
- Adds `sign(payload)` to the `ApiKeyStamper` type, plumbing through to the underlying IndexedDB stamper's existing `sign` method.
- `authenticateWithOAuth` now accepts a required `popSignature` and includes it in the `POST /{projectId}/auth/oauth` body.
- `createZeroDevWallet.auth({ type: 'oauth' })` signs the incoming `sessionId` with the active session key and forwards the signature as `popSignature`.

Coordinates with the corresponding backend change. Smoke-tested locally end-to-end — Google login succeeds and the request body carries both `sessionId` and `popSignature`.

## Notes
- No public API change at the React layer — signing happens inside `wallet.auth`, so `authenticateOAuth` / `useAuthenticateOAuth` callers are unaffected.
- Backend gracefully ignores the extra field on older deployments (Gin's default JSON binding drops unknown keys), so this SDK can ship before the backend rolls out.
- Changeset to follow (will run `pnpm changeset` interactively).

### DIFF
--- a/packages/core/src/actions/auth/auth.test.ts
+++ b/packages/core/src/actions/auth/auth.test.ts
@@ -55,7 +55,7 @@ function createMockClient(
 }
 
 describe('authenticateWithOAuth', () => {
-  it('sends OAuth auth request with sessionId in body', async () => {
+  it('sends OAuth auth request with sessionId and popSignature in body', async () => {
     const mockClient = createMockClient(async () => ({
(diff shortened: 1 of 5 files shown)


andrascodes (approved):
Very cool!! lgtm

PR #233, opened 2026-05-26
fix: split web and native wallet connectors

andrascodes:

* **#234**
* ➡️ **#233**
* `main`
</details></td></tr></table>

Closes FS-2252

## Summary

- Split the React connector into shared core logic plus web/native wrappers.
- Add React Native defaults that derive `Origin: https://${rpId}` from the connector config.
- Remove `rpId` as a required prop from `ZeroDevExportWebView`; it now reads it from the ZeroDev wagmi connector.
- Add a native core SDK wrapper that applies the same Turnkey Origin default.
- Add separate Expo native/web typecheck configs.

## Why

The connector now has platform-specific API needs. On React Native, consumers must provide native adapters and the SDK must send an `Origin` header that matches `rpId` so Turnkey and AA requests pass ACL checks. On web, browsers own the `Origin` header, so exposing `fetchOptions` for this path is misleading and does not solve the same problem.

Splitting the connector keeps the public API accurate for each platform while preserving one shared implementation for the actual wagmi connector behavior. The native wrapper can require the RN-only adapter fields and derive the Origin header automatically, while the web wrapper can keep the browser API smaller and avoid exposing options that do not apply there.

If Passkeys are used `rpId` needs to be a domain associated with the app - the KMS also enforces that the Passkey requests come from the same domain. So it makes sense to default the Origin header and the Passkey `rpId` to the same domain and have 1 source of truth.
- When developing locally on a ZD project where an Origin allowlist is defined the developer can pass the localhost URL (for example `http://localhost:3000`) as the `rpId` and that will work for basic KMS requests (OTP email, Google OAuth). To use Passkeys or OTP Magic Link the user needs to set up Universal Links (with a domain association) and then the developer should use the same domain as the `rpId` as well. This way requests coming from the app, on the KMS's side will look like they're coming from the associated domain (`Origin` is set to the domain).

## Testing

- Tested the `zerodev-signer-demo` on web, the `expo-example` on both web and Android. All working.


packages/core/src/stubs/native-on-web.ts
```diff
@@ -1,3 +1,25 @@
-throw new Error(
```
OmarBasem, line 1: what do you mean by `native-on-web` is it `react-native-web` ? If so, I think naming it `react-native-web.ts` would be more clear

packages/core/src/utils/originFromRpId.ts
```diff
@@ -0,0 +1,12 @@
...
+ * `rpId`. If the value already includes an `http://` or `https://` scheme,
+ * return it verbatim — useful in dev when a backend ACL requires a
+ * non-https origin like `http://localhost:8082`. Otherwise prepend
+ * `https://`. Internal — not re-exported from the package.
+ */
+export function originFromRpId(rpId: string): string {
+  return rpId.startsWith('http://') || rpId.startsWith('https://')
+    ? rpId
+    : `https://${rpId}`
+}
```
spsjvc: Should we ever allow http without https? Maybe only if it's localhost, otherwise throw an error. Also, origin shouldn't have subpaths, e.g. `https://zerodev.app/example` and we currently just using it as-is.

And a nit, thinking we can rename this to `sanitizeOrigin` or `normalizeOrigin` since it's applicable more broadly.

packages/react/src/utils/originFromRpId.ts
spsjvc, line 1: Isn't this exactly the same as `packages/core/src/utils/originFromRpId.ts`?

packages/core/src/stubs/native-on-web.ts
andrascodes: This has nothing to do with `react-native-web`. It's a file that is loaded when React Native code is imported on a Web environment ~and it throws an error.~
Correction: it throws when the imported functions are called on the wrong platform. Not on import. This is documented in the comments in the file.

packages/react/src/utils/originFromRpId.ts
andrascodes: yes, I'll change it so we import the one from `wallet-core` instead

packages/core/src/utils/originFromRpId.ts
andrascodes: reverted the file

spsjvc (approved):

PR #234, opened 2026-05-28
feat: make passkeyStamper optional

andrascodes:
Closes FS-2252, FS-2251

If users don't need passkey sign-in they shouldn't need to pass a `passkeyStamper` on React Native. This PR makes the `passkeyStamper` param optional on the `zeroDevWallet` connector and `createZeroDevWallet` function.
I added a `noopPasskeyStamper` which is passed inside the SDK if no `passkeyStamper` is passed from the configs. This throws a descriptive error asking the user to pass a real stamper. This way we also didn't need to make any modifications to the underlying code.

### DIFF
--- a/apps/expo-example/docs/QUICK_START.md
+++ b/apps/expo-example/docs/QUICK_START.md
@@ -1,0 +1,323 @@
+# Quick Start (Expo / React Native)
+
+Stand up a React Native app that signs in with the ZeroDev wallet and
+shows the connected address. The path here uses **email OTP** auth plus
+an optional **Google OAuth** flow — neither needs a verified domain or
+`assetlinks.json`, so you can run end-to-end with nothing more than a
+project ID. Passkeys are out of scope for this guide (they require domain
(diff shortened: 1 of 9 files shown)


apps/expo-example/docs/QUICK_START.md
```diff
@@ -0,0 +1,323 @@
+# Quick Start (Expo / React Native)
```
OmarBasem, line 1: I think we can add a section about `react-native-web` to the docs, and I am assuming stampers usage/setup would be more similar to plain react so we can explain that in the docs.

spsjvc (approved):

apps/expo-example/docs/QUICK_START.md
andrascodes: Thanks! I'll add this later on the actual docs. This Quick start is only meant to unblock a team quickly that might need the RN SDK before I can complete the ZD docs changes.

PR #238, opened 2026-06-02
fix(react-kit): url redirect

OmarBasem:


brtkx (approved):

packages/react-kit/src/connector.ts
```diff
@@ -61,20 +61,39 @@ export function zeroDevWallet(
...
+        // wagmi sets `isReconnecting` on reconnectOnMount.
+        const isReconnecting = !!connectParams && connectParams.isReconnecting
+
+        // Force the base connector to finish initializing so its store can
+        // tell us if a session is already persisted. Base `isAuthorized()`
+        // only reads `store.eoaAccount` synchronously and reports `false`
+        // before init completes — surfacing SignUp here causes a visible
+        // flash on the next paint when a session exists but isn't loaded
+        // yet (common on first paint of a production build).
+        if ('getStore' in connector && typeof connector.getStore === 'function')
```
brtkx: I'm not sure if this is a good idea. It is avoided in the core package as well because it would always await for the store initialization:
https://github.com/zerodevapp/zerodev-wallet-sdk/blob/main/packages/react/src/core/connector.ts#L401

What if we tried something like this? It may not be the exact solution but just a whiteboard example. What it does it's trying to connect with the connector first, and if it fails then we do some checks without having to *always* await the store.
```typescript
function isNotAuthenticatedError(error: unknown): boolean {
  return (
    error instanceof Error && error.message.toLowerCase().includes('not authenticated')
  )
}

function waitForAuthFlow(store: ReturnType<typeof createStore>): Promise<void> {
  return new Promise<void>((resolve, reject) => {
    const unsub = store.subscribe(
      (state) => state.auth.step,
      (step) => {
        if (step === 'authenticated') {
          unsub()
          resolve()
        } else if (step === null) {
          unsub()
          reject(new Error('Auth flow dismissed'))
        }
      },
    )
  })
}

      async connect(connectParams) {
        try {
          return await connector.connect(connectParams)
        } catch (error) {
          const isReconnecting =
            !!connectParams && connectParams.isReconnecting === true
          if (
            isReconnecting ||
            !params.config?.auth ||
            store.getState().auth.step !== null ||
            !isNotAuthenticatedError(error)
          ) {
            throw error
          }

          store.getState().auth.goToStep('sign-up')
          await waitForAuthFlow(store)
          return connector.connect(connectParams)
        }
```

brtkx (dismissed):

packages/react-kit/src/connector.ts
OmarBasem: @brtkx refactored, and added a `NotAuthenticatedError` to `react` package for a more reliable and type-safe error matching instead of comparing substring

packages/react-kit/src/connector.ts
```diff
@@ -61,26 +87,35 @@ export function zeroDevWallet(
...
+        // (which `isAuthorized()` deliberately skips for perf — see
+        // packages/react/src/core/connector.ts), and avoids flashing the auth
+        // UI on first paint when the session hasn't loaded yet.
+        try {
+          return await connector.connect(connectParams)
+        } catch (error) {
+          // wagmi sets `isReconnecting` on reconnectOnMount; a silent
+          // reconnect failure must not pop the auth UI.
+          const isReconnecting =
+            !!connectParams && connectParams.isReconnecting === true
```
brtkx: nit: this should be enough?
```typescript
!!connectParams && connectParams.isReconnecting
```

packages/react-kit/src/connector.ts
```diff
@@ -61,26 +87,35 @@ export function zeroDevWallet(
...
+
+          if (
```
brtkx, line 101: Maybe we should do this? The error is more intentional, otherwise if a dev triggers the connection while sign-up is in progress, they will get an unauthenticated error from react package.
```typescript
if (
  isReconnecting ||
  !params.config?.auth ||
  !(error instanceof NotAuthenticatedError)
) {
  throw error
}

if (store.getState().auth.step !== null) {
  throw new Error('Auth flow already in progress')
}
```

packages/react-kit/src/connector.ts
OmarBasem: refactored to `!!connectParams?.isReconnecting`

packages/react-kit/src/connector.ts
OmarBasem: updated

packages/react/src/index.ts
```diff
@@ -12,6 +12,7 @@ export {
   getZeroDevWallet,
 } from './actions.js'
 export type { GetOAuthSessionIdFn } from './authenticateOAuth.js'
+export { NotAuthenticatedError } from './errors.js'
```
brtkx, line 15: I think we should also add `export { NotAuthenticatedError } from './errors.js'` to `index.native.ts`

brtkx (approved):

PR #240, opened 2026-06-05
feat(react-kit): add `useSmartRoutingAddress` hook + connector config

OmarBasem:
Closes [FS-2261](https://linear.app/offchain-labs/issue/FS-2261/connector-config-usesmartroutingaddress-hook)

### Summary

Foundation for the SRA UI, no components in this PR.

-  Adds `@zerodev/smart-routing-address` and a new smartRoutingAddress block on zeroDevWallet({ config }) (enabled, maxSlippage, destinationChains, sourceChains).
- Adds a smartRouting slice to the kit store.  
-  Adds `useSmartRoutingAddress(overrides?)` wraps createSmartRoutingAddress in React Query

### DIFF
--- a/packages/react-kit/src/smart-routing/useSmartRoutingAddress.test.tsx
+++ b/packages/react-kit/src/smart-routing/useSmartRoutingAddress.test.tsx
@@ -36,15 +36,9 @@ const mockConfig = {
     getKitStore?: () => unknown
   }>,
 }
-let mockChainId = mainnet.id
-const mockChains = [mainnet, optimism]
-const mockAccount = { address: WALLET as `0x${string}` | undefined }
 
 vi.mock('wagmi', () => ({
   useConfig: () => mockConfig,
-  useAccount: () => mockAccount,
-  useChainId: () => mockChainId,
-  useChains: () => mockChains,
 }))
 
 function makeWrapper() {
@@ -56,11 +50,15 @@ function makeWrapper() {
   )
 }
 
+const baseParams = {
+  owner: WALLET,
+  destChain: mainnet,
+  srcTokens: [{ tokenType: 'USDC' as const, chain: optimism }],
+}
+
 beforeEach(() => {
   mockCreateSmartRoutingAddress.mockReset()
   mockCreateCall.mockClear()
-  mockChainId = mainnet.id
-  mockAccount.address = WALLET
   mockConfig.connectors = [mockConnector]
   mockConnector.getKitStore.mockReset()
 })
@@ -74,7 +72,7 @@ describe('useSmartRoutingAddress', () => {
     mockConfig.connectors = []
     const wrapper = makeWrapper()
     expect(() =>
-      renderHook(() => useSmartRoutingAddress(), { wrapper }),
+      renderHook(() => useSmartRoutingAddress(baseParams), { wrapper }),
     ).toThrow(
       /useSmartRoutingAddress must be used with zeroDevWallet connector/,
     )
@@ -85,13 +83,7 @@ describe('useSmartRoutingAddress', () => {
     mockConnector.getKitStore.mockReturnValue(store)
 
     const wrapper = makeWrapper()
-    renderHook(
-      () =>
-        useSmartRoutingAddress({
-          srcTokens: [{ tokenType: 'ERC20', chain: optimism }],
-        }),
-      { wrapper },
-    )
+    renderHook(() => useSmartRoutingAddress(baseParams), { wrapper })
 
     expect(mockCreateSmartRoutingAddress).not.toHaveBeenCalled()
   })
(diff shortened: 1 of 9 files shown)


packages/react-kit/src/smart-routing/useSmartRoutingAddress.ts
```diff
@@ -0,0 +1,125 @@
...
+    store,
+    (state) => state.smartRouting.config,
+  ) as SmartRoutingAddressConfig | null
+
+  const { address: connectedAddress } = useAccount()
+  const currentChainId = useChainId()
+  const availableChains = useChains()
+
+  const owner = overrides.owner ?? connectedAddress
+  const destChain =
```
brtkx: Does the hook need to figure out the defaults at all? Couldn't we just make destination chain a required parameter of the hook? 

What would be a situation where the consumer doesn't need to provide the destination chain, and the hook needs to default it? If there's no such scenario, we can just simplify it.

(applies to all overrides)

OmarBasem: ~I agree the hook doesn't need to figure out the defaults. I updated in [that PR](https://github.com/zerodevapp/zerodev-wallet-sdk/pull/244/) to avoid conflicts, the file would now look [as follows](https://github.com/zerodevapp/zerodev-wallet-sdk/blob/4401251d638e6b531a62cc44ef7e41b651537c29/packages/react-kit/src/smart-routing/hooks/useSmartRoutingAddress.ts)~

packages/react-kit/src/smart-routing/useSmartRoutingAddress.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { mainnet, optimism } from 'viem/chains'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { createStore } from '../store'
+import { useSmartRoutingAddress } from './useSmartRoutingAddress'
+
+const mockCreateSmartRoutingAddress = vi.fn()
+const mockCreateCall = vi.fn((args) => ({ __call: args }))
+
+vi.mock('@zerodev/smart-routing-address', () => ({
```
brtkx: Would this work? We should aim to only mock what's necessary. Otherwise our tests become circular and possibly give false positives because we assert a function that's just mocked.
```typescript
vi.mock('@zerodev/smart-routing-address', async () => {
  const actual =
    await vi.importActual<typeof import('@zerodev/smart-routing-address')>(
      '@zerodev/smart-routing-address',
    )

  return {
    ...actual,
    createSmartRoutingAddress: (params: unknown) =>
      mockCreateSmartRoutingAddress(params),
  }
})
```

packages/react-kit/src/smart-routing/useSmartRoutingAddress.ts
OmarBasem: @brtkx updated `useSmartRoutingAddress` in the current PR

packages/react-kit/src/smart-routing/useSmartRoutingAddress.test.tsx
OmarBasem: Updated

packages/react-kit/src/smart-routing/useSmartRoutingAddress.ts
```diff
@@ -0,0 +1,110 @@
...
+  const resolvedActions = actions ?? buildDefaultActions(owner, srcTokens)
+
+  return useQuery({
+    queryKey: [
+      'zerodev-wallet-kit',
+      'smartRoutingAddress',
+      owner,
+      destChain.id,
+      slippage,
+      srcTokens.map((t) => [t.tokenType, t.chain.id, t.minAmount?.toString()]),
```
brtkx, line 86: Need to add `actions` to the query key?

packages/react-kit/src/smart-routing/useSmartRoutingAddress.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * The query is disabled when the SRA feature flag is off
+ * (`config.smartRoutingAddress.enabled === false`).
+ */
+export function useSmartRoutingAddress({
+  owner,
+  destChain,
+  srcTokens,
+  slippage = DEFAULT_SLIPPAGE_BPS,
+  actions,
```
brtkx: Is there a reason why `actions` is exposed as param? I don't know what the scenario would be for actions to be something custom other than the transfer of funds. 

If it's not intended, then it's a bit of a speculative abstraction and we could remove `actions` from here. Then, we could also rename `buildDefaultActions` to `buildActions`.

(if we do that, then [this](https://github.com/zerodevapp/zerodev-wallet-sdk/pull/240#discussion_r3388732483) comment doesn't apply)

OmarBasem: good catch, it was not intended

brtkx (dismissed):

PR #241, opened 2026-06-08
feat(react-kit): Select component

OmarBasem:
Closes [FS-2267](https://linear.app/offchain-labs/issue/FS-2267/select-component)

### Summary

This PR implements `Select` component

### DIFF
--- a/packages/react-kit/src/shared/components/Select/index.tsx
+++ b/packages/react-kit/src/shared/components/Select/index.tsx
@@ -1,0 +1,107 @@
+import type { ButtonHTMLAttributes } from 'react'
+
+import { cn } from '../../utils/common'
+import { Icon } from '../Icon'
+import { Text } from '../Text'
+import { Wrapper } from '../Wrapper'
+
+export interface SelectProps
+  extends Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'children'> {
+  /** Primary label (e.g. token symbol). */
+  label: string
+  /** Optional subtitle row (e.g. chain name). */
+  subtitle?: string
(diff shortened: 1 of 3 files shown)


packages/react-kit/src/shared/components/Select/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { Icon } from '../Icon'
+import { Text } from '../Text'
+import { Wrapper } from '../Wrapper'
+
+export interface SelectProps
+  extends Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'children'> {
+  /** Primary label (e.g. token symbol). */
+  label: string
+  /** Optional subtitle row (e.g. chain name). */
+  subtitle?: string
```
brtkx, line 13: 1. Are `subtitle` and `chainImage` going to be used anywhere for SRA?
2. Maybe we should rename `chainImage` to `trailingImage` to make it consistent with being content-agnostic?

OmarBasem: `subtitle` and `chainImage` are used in the `Select` component in [other pages](https://www.figma.com/design/MrTOqWtVg3xPNXFJLDB9m6/Doorway-Design-System?node-id=789-17170&m=dev)

I think using `trailingImage` here wouldn't be appropriate, if you look at the [design here](https://www.figma.com/design/MrTOqWtVg3xPNXFJLDB9m6/Doorway-Design-System?node-id=789-17170&m=dev) you would see it's not really a trailing image

brtkx (approved):

PR #245, opened 2026-06-08
feat(react): support wallet_sendCalls in the provider

brtkx:

### DIFF
--- a/packages/react/src/provider.ts
+++ b/packages/react/src/provider.ts
@@ -11,6 +11,12 @@ const SESSION_WARNING_THRESHOLD_MS = 60 * 1000 // 1 minute before expiry
 
 type Signer = SmartAccount<KernelSmartAccountImplementation> | LocalAccount
 
+type WalletCall = {
+  to?: `0x${string}`
+  value?: `0x${string}`
+  data?: `0x${string}`
+}
+
 /**
  * Pick the right signer for `personal_sign` / `eth_signTypedData_v4`.
  *
@@ -256,20 +262,84 @@ export function createProvider({
             throw new Error('wallet_sendCalls is not supported in EOA mode')
           }
 
+          if (!Array.isArray(request.calls) || request.calls.length === 0) {
+            throw new Error('Missing calls')
+          }
+
           const kernelClient = store.getState().kernelClients.get(chainId)
           if (!kernelClient) {
             throw new Error(`No kernel client for chain ${chainId}`)
           }
 
-          // TODO: Type the provider RPC params as a shared request union so
-          // `request.calls` can be narrowed without `any`.
-          return await kernelClient.sendTransaction({
-            calls: request.calls.map((call: any) => ({
+          // EIP-5792 is async by design: return the bundle ID immediately,
+          // dapps poll wallet_getCallsStatus. The userOp hash is the ID.
+          const userOpHash = await kernelClient.sendUserOperation({
+            calls: request.calls.map((call: WalletCall) => ({
               ...(call.to && { to: call.to }),
               value: call.value ? BigInt(call.value) : 0n,
               data: call.data || '0x',
             })),
           })
+          return { id: userOpHash }
+        }
+
+        case 'wallet_getCallsStatus': {
+          if (!params || params.length === 0) {
+            throw new Error('Missing call bundle id')
+          }
+
+          const [id] = params
+          const kernelClient = store.getState().kernelClients.get(activeChainId)
+          if (!kernelClient) {
+            throw new Error(`No kernel client for chain ${activeChainId}`)
+          }
+
+          const receipt = await kernelClient
+            .getUserOperationReceipt({ hash: id })
+            .catch(() => null)
+          if (!receipt) {
+            return { version: '2.0.0', atomic: true, status: 100, receipts: [] }
+          }
+
+          // viem's getCallsStatus decoder expects RPC-hex receipts (it runs
+          // hexToBigInt on blockNumber/gasUsed), but the bundler client
+          // returns parsed bigint fields — re-encode them.
+          const r = receipt.receipt
+          return {
+            version: '2.0.0',
+            atomic: true,
+            chainId: `0x${activeChainId.toString(16)}`,
+            status: receipt.success ? 200 : 500,
+            receipts: [
+              {
+                transactionHash: r.transactionHash,
+                blockHash: r.blockHash,
+                blockNumber: `0x${r.blockNumber.toString(16)}`,
+                gasUsed: `0x${r.gasUsed.toString(16)}`,
+                status: r.status === 'success' ? '0x1' : '0x0',
+                // receipt.logs is scoped to this userOp — r.logs would leak
+                // logs from other userOps bundled into the same transaction.
+                logs: receipt.logs.map((log) => ({
+                  address: log.address,
+                  topics: log.topics,
+                  data: log.data,
+                })),
+              },
+            ],
+          }
+        }
+
+        case 'wallet_getCapabilities': {
+          const [, requestedChainIds] = params || []
+          const chains: string[] = requestedChainIds ?? [
+            `0x${activeChainId.toString(16)}`,
+          ]
+          // Sponsorship is project-level (dashboard policies / Ultra Relay),
+          // so we intentionally do not declare or honor dapp-supplied
+          // paymasterService.
+          const atomic =
+            mode === 'EOA' ? { status: 'unsupported' } : { status: 'supported' }
+          return Object.fromEntries(chains.map((cid) => [cid, { atomic }]))
         }
 
         case 'personal_sign': {
--- a/packages/react/src/provider.ts
+++ b/packages/react/src/provider.ts
@@ -276,7 +276,8 @@ export function createProvider({
           const userOpHash = await kernelClient.sendUserOperation({
             calls: request.calls.map((call: WalletCall) => ({
               ...(call.to && { to: call.to }),
-              value: call.value ? BigInt(call.value) : 0n,
+              value:
+                call.value && call.value !== '0x' ? BigInt(call.value) : 0n,
               data: call.data || '0x',
             })),
           })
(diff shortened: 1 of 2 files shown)


packages/react/src/provider.ts
```diff
@@ -223,6 +240,37 @@ export function createProvider({
...
+
+          // TODO: Type the provider RPC params as a shared request union so
+          // `request.calls` can be narrowed without `any`.
+          return await kernelClient.sendTransaction({
+            calls: request.calls.map((call: any) => ({
+              ...(call.to && { to: call.to }),
+              value: call.value ? BigInt(call.value) : 0n,
+              data: call.data || '0x',
+            })),
+          })
```
SahilVasava: Two issues with `sendTransaction` here (verified against `@zerodev/sdk@5.5.4` — `actions/account-client/sendTransaction.js`):

1. **It blocks until inclusion.** Internally it does `sendUserOperation` → `waitForUserOperationReceipt` → returns the receipt's `transactionHash`. So this RPC call hangs for the full inclusion time. EIP-5792 is async by design: submit → return ID immediately → dapp polls `wallet_getCallsStatus`.
2. **It returns the wrong identifier.** The tx hash comes back, but the natural call-bundle ID in the AA world is the **userOp hash** — that's what `wallet_getCallsStatus` would key on.

Switching to `sendUserOperation` fixes both, and wrapping as `{ id }` matches the spec shape (viem tolerates a bare string and wraps it — `sendCalls.ts:147` — but `{ id }` is the documented contract):

```suggestion
          const userOpHash = await kernelClient.sendUserOperation({
            calls: request.calls.map((call: any) => ({
              ...(call.to && { to: call.to }),
              value: call.value ? BigInt(call.value) : 0n,
              data: call.data || '0x',
            })),
          })
          return { id: userOpHash }
```

packages/react/src/provider.ts
```diff
@@ -223,6 +240,37 @@ export function createProvider({
...
+        }
```
SahilVasava, line 348: `wallet_sendCalls` needs its polling counterpart, `wallet_getCallsStatus` — viem's `getCallsStatus` action / wagmi's `useCallsStatus` call it directly with the ID, and without it consumers hang on "pending" forever.

Two gotchas from viem 2.38.6's decoder (`actions/wallet/getCallsStatus.ts`):
- `status` is **numeric** in the current spec: `100` pending, `200` success, `500` failure (the `'PENDING'`/`'CONFIRMED'` strings are legacy back-compat).
- It post-processes receipts with `hexToBigInt(receipt.blockNumber)` / `hexToBigInt(receipt.gasUsed)` — so receipts must go back in **RPC hex format**. `kernelClient.getUserOperationReceipt` returns *parsed* receipts (bigint fields), so they need re-encoding:

```ts
case 'wallet_getCallsStatus': {
  const [id] = params
  const kernelClient = store.getState().kernelClients.get(activeChainId)
  if (!kernelClient) throw new Error(`No kernel client for chain ${activeChainId}`)
  const receipt = await kernelClient
    .getUserOperationReceipt({ hash: id })
    .catch(() => null)
  if (!receipt)
    return { version: '2.0.0', atomic: true, status: 100, receipts: [] }
  const r = receipt.receipt
  return {
    version: '2.0.0',
    atomic: true,
    chainId: `0x${activeChainId.toString(16)}`,
    status: receipt.success ? 200 : 500,
    receipts: [
      {
        transactionHash: r.transactionHash,
        blockHash: r.blockHash,
        blockNumber: `0x${r.blockNumber.toString(16)}`,
        gasUsed: `0x${r.gasUsed.toString(16)}`,
        status: r.status === 'success' ? '0x1' : '0x0',
        logs: r.logs,
      },
    ],
  }
}
```

packages/react/src/provider.ts
```diff
@@ -223,6 +240,37 @@ export function createProvider({
           })
         }
 
+        case 'wallet_sendCalls': {
```
SahilVasava, line 250: Suggest also adding a `wallet_getCapabilities` case in this PR — dapps probe it before deciding whether to use `sendCalls` at all (viem's `getCapabilities` action hits it directly; params shape is `[address, [chainIdHex]?]`). Without it, capability-aware dapps silently fall back to legacy `eth_sendTransaction` and never exercise this code path.

Declare what we actually support — `atomic` when a kernel is backing the account (`executeBatch` is atomic), `unsupported` in EOA mode since `wallet_sendCalls` throws there. **Deliberately don't declare `paymasterService`**: in our model sponsorship is configured at the project level (ZD dashboard policies / Ultra Relay auto-sponsorship), not accepted from the dapp at call time. Worth a code comment saying so, since ignoring dapp-supplied `capabilities.paymasterService` is intentional.

```ts
case 'wallet_getCapabilities': {
  const [, requestedChainIds] = params || []
  const chains: string[] =
    requestedChainIds ?? [`0x${activeChainId.toString(16)}`]
  // Sponsorship is project-level (dashboard policies / Ultra Relay), so we
  // intentionally do not declare or honor dapp-supplied paymasterService.
  const atomic =
    mode === 'EOA' ? { status: 'unsupported' } : { status: 'supported' }
  return Object.fromEntries(chains.map((cid) => [cid, { atomic }]))
}
```

Dapps seeing `unsupported` fall back to sequential `eth_sendTransaction`, which is exactly right for EOA mode.

packages/react/src/provider.ts
```diff
@@ -223,6 +240,37 @@ export function createProvider({
...
+
+          const kernelClient = store.getState().kernelClients.get(chainId)
+          if (!kernelClient) {
+            throw new Error(`No kernel client for chain ${chainId}`)
+          }
+
+          // TODO: Type the provider RPC params as a shared request union so
+          // `request.calls` can be narrowed without `any`.
+          return await kernelClient.sendTransaction({
+            calls: request.calls.map((call: any) => ({
```
SahilVasava: `request.calls.map(...)` throws a raw `TypeError` if `calls` is missing or not an array. A guard keeps the error consistent with the rest of this file's defensive style:

```ts
if (!Array.isArray(request.calls) || request.calls.length === 0) {
  throw new Error('Missing calls')
}
```

packages/react/src/provider.ts
```diff
@@ -223,6 +240,37 @@ export function createProvider({
...
+            throw new Error('wallet_sendCalls is not supported in EOA mode')
+          }
+
+          const kernelClient = store.getState().kernelClients.get(chainId)
+          if (!kernelClient) {
+            throw new Error(`No kernel client for chain ${chainId}`)
+          }
+
+          // TODO: Type the provider RPC params as a shared request union so
+          // `request.calls` can be narrowed without `any`.
```
SahilVasava: The union here is small enough to just type inline now rather than TODO it:

```ts
type WalletCall = {
  to?: `0x${string}`
  value?: `0x${string}`
  data?: `0x${string}`
}
```

then `request.calls.map((call: WalletCall) => ...)` and the `any` disappears.

SahilVasava (commented):
Solid foundation — the `accountAddressForChain` refactor and `validateFromAddress` are clean wins, and rejecting EOA mode is correct.

Main theme of the inline comments: as written this is a partial EIP-5792 — the return shape/timing of `wallet_sendCalls` diverges from the spec, and the two companion methods dapps probe/poll (`wallet_getCapabilities`, `wallet_getCallsStatus`) are missing. Details inline; all verified against viem 2.38.6 and @zerodev/sdk 5.5.4 sources.

Also: no tests cover the new case. Worth adding at least: happy-path `wallet_sendCalls`, EOA-mode reject, mismatched `from` throw, and missing/malformed `calls`.

(`wallet_showCallsStatus` — the 4th 5792 method — is intentionally fine to skip: it asks the wallet to show its own status UI, which doesn't apply to a headless embedded wallet.)

packages/react/src/provider.ts
brtkx: done

packages/react/src/provider.ts
brtkx: added

packages/react/src/provider.ts
```diff
@@ -224,6 +247,101 @@ export function createProvider({
...
+          return { id: userOpHash }
+        }
+
+        case 'wallet_getCallsStatus': {
+          if (!params || params.length === 0) {
+            throw new Error('Missing call bundle id')
+          }
+
+          const [id] = params
+          const kernelClient = store.getState().kernelClients.get(activeChainId)
```
SahilVasava: Non-blocking, but worth a note: this resolves the kernel client by `activeChainId`, whereas `wallet_sendCalls` submits on `request.chainId`. If a dapp passes an explicit `chainId` that differs from the active chain and then polls status, this queries the active chain's bundler — which never saw that userOp — so `getUserOperationReceipt` returns null and the bundle stays `pending` (100) forever.

Narrow in practice (wagmi switches the active chain before sending, so they usually match), but it's a latent correctness bug. Since EIP-5792 ids are opaque and viem passes them straight through to `wallet_getCallsStatus`, the clean fix is to encode the chain into the id:

```ts
// wallet_sendCalls
return { id: `${userOpHash}:${chainId}` }

// wallet_getCallsStatus
const [rawId] = params
const [hash, encodedChainId] = rawId.split(':')
const chainId = encodedChainId ? Number(encodedChainId) : activeChainId
const kernelClient = store.getState().kernelClients.get(chainId)
```

Or just leave a comment documenting the active-chain assumption if cross-chain `sendCalls` isn't a flow you support yet.

SahilVasava (approved):
All five points from the last round are addressed — and the `wallet_getCallsStatus` receipt handling is actually better than I suggested: using `receipt.logs` (scoped to the userOp) rather than `receipt.receipt.logs` avoids leaking logs from sibling userOps bundled into the same tx. Verified against viem's `UserOperationReceipt` type.

Ran it locally: 12 new tests pass, react typecheck + core build clean, and the 100/200/500 status codes match viem's `getCallsStatus` decoder for atomic batches.

Approving. One non-blocking edge case inline that's worth a follow-up but doesn't need to hold this.

PR #248, opened 2026-06-10
Update signer demo auth layout

srinjoyc:
## Summary

Updates the signer demo auth layout so the wallet panel is smaller and positioned higher on desktop.

Also moves the email auth settings control into the auth panel's top-right area and hides the close button from the example app only. No React kit package source changes are included.


srinjoyc:
Superseded by #249 using the renamed srinjoy/demo-auth-layout-updates branch.

PR #255, opened 2026-06-16
fix(core): sign stamp-login against the parent org

SahilVasava:
## What

Coordinates with the doorway-kms backend change (OffchainLabs/doorway-kms#778) that removes the client-supplied sub-org id from `/auth/login/stamp`. The backend now relays the stamp-login to Turnkey under the **parent (base) org** and derives the user's sub-org from the stamped credential.

Because the X-Stamp signs a payload that includes the org id, the SDK must now sign against the **parent** org. Signing the sub-org makes the org in the relayed payload mismatch the signature, and Turnkey rejects it with `SIGNATURE_INVALID`.

## Changes

- `loginWithStamp` callsites in `createZeroDevWalletCore` — register bootstrap, session refresh, and passkey login — now sign `organizationId` (the connector's configured parent org) instead of the per-user sub-org. Passkey login already did; register-bootstrap and refresh were signing the sub-org and are the ones that broke.
- `loginWithStamp` no longer sends `subOrganizationId` in the request body — the backend derives it from the credential.
- The parent org is the connector's existing `organizationId` default, so there's **no extra round-trip** (the new `/server-info/parent-org-id` endpoint is not needed by the SDK).

## Verification

Ran the current SDK against a local backend on the #778 branch:
- **Before:** passkey register and email/OTP session refresh both fail — backend logs `StampLogin … SIGNATURE_INVALID` against the parent org.
- **After:** passkey register (browser e2e, virtual authenticator) and email session refresh (`session-management` integration test) both succeed.

Tests:
- `auth.test.ts` — asserts the stamp signs the parent and the body no longer carries `subOrganizationId`.
- `session-management.test.ts` — OTP login → refresh signs the parent (via new `getParentOrgId` helper) → whoami on the rotated session.

## Rollout

Coordinated, no compat window: the backend unconditionally stamps the parent, so this SDK must ship before #778 deploys.

### DIFF
--- a/apps/zerodev-signer-demo/src/app/wagmi-config.ts
+++ b/apps/zerodev-signer-demo/src/app/wagmi-config.ts
@@ -32,6 +32,11 @@ export const config = createConfig({
       projectId: process.env.NEXT_PUBLIC_ZERODEV_PROJECT_ID!,
       proxyBaseUrl: process.env.NEXT_PUBLIC_KMS_PROXY_BASE_URL!,
       chains: [arbitrumSepolia, sepolia],
+      // Local testing override: our docker backend's Turnkey base org differs
+      // from the SDK's hardcoded prod default, so point the connector at it.
+      ...(process.env.NEXT_PUBLIC_ZERODEV_ORG_ID && {
+        organizationId: process.env.NEXT_PUBLIC_ZERODEV_ORG_ID,
+      }),
       ...(mode && { mode }),
       config: {
         auth: {
--- a/e2e/helpers/backend-health.ts
+++ b/e2e/helpers/backend-health.ts
@@ -53,3 +53,20 @@ export async function getAuthProxyConfigId(baseUrl: string): Promise<string> {
   const data = await res.json()
   return data.authProxyConfigId
 }
+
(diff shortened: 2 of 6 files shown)


apps/zerodev-signer-demo/src/app/wagmi-config.ts
```diff
@@ -32,6 +32,11 @@ export const config = createConfig({
       projectId: process.env.NEXT_PUBLIC_ZERODEV_PROJECT_ID!,
       proxyBaseUrl: process.env.NEXT_PUBLIC_KMS_PROXY_BASE_URL!,
       chains: [arbitrumSepolia, sepolia],
+      // Local testing override: our docker backend's Turnkey base org differs
+      // from the SDK's hardcoded prod default, so point the connector at it.
```
andrascodes, line 36: Is this okay to leave in or did you forget about it and it was only used temporarily for testing?

andrascodes: nvm, I understand now

andrascodes (approved):
LGTM

PR #273, opened 2026-06-30
feat(monorepo): add @zerodev/react-ui and @zerodev/react-wallet-ui packages

brtkx:
## Summary

Adds two new workspace packages: `@zerodev/react-ui` (branded UI primitives) and `@zerodev/react-wallet-ui`. Additive — existing `react-kit` is untouched.

## Changes

- `react-ui` and `react-wallet-ui` from `react-monorepo` moved exactly as they were. The review focus should be on the folder structure being the same as in the monorepo. Code is the same, with the following minor adjustments:
- `pnpm-lock.yaml` re-resolved for the two new packages (the 680 deletions are lockfile churn only — no source removed).
- Both packages are `private: true` at `0.0.1-alpha.0` — not yet published.
- `react-kit` stays as is - will remove later.
- ZeroDev demo running with `react-kit` - will make it run with `react-wallet-ui` in the next PRs.
- Adjusted typescript config to point at the base at `/templates`.


.github/workflows/ci.yml
```diff
@@ -95,6 +95,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build packages
+        run: pnpm build
+
```
OmarBasem: Curious why do we need this now if we didn't have it before

packages/react-wallet-ui/src/shared/components/PoweredBy/index.tsx
OmarBasem, line 1: I think maybe we don't need `shared` directory under `react-wallet-ui` now that we have `react-ui` that is supposed to house the reusable primitives and components?

brtkx: `react-wallet-ui` can still have components that are shared within this package only. It shouldn't automatically be part of `react-ui` just because it's shared here.

A component should earn its place to be part of `react-ui`. E.g.
- a primitive that is going to be used across different packages, such as button, input, checkbox.
- a component that we are reusing in 3+ packages; e.g. if `PoweredBy` was used consistently in other UI packages, then we could consider making it part of `react-ui`, but we shouldn't do early abstractions imo

.github/workflows/ci.yml
brtkx: it's because it's a dependency of `react-wallet-ui`; but you're right we don't need it. I had to add redirect in vitest config instead. Fixed

OmarBasem (approved):

PR #280, opened 2026-07-02
chore: upgrade expo to SDK 56 and fix oauth

OmarBasem:


package.json
```diff
@@ -55,6 +55,11 @@
     "vitest": "^4.0.18"
   },
   "packageManager": "pnpm@10.17.1+sha512.17c560fca4867ae9473a3899ad84a88334914f379be46d455cbf92e5cf4b39d34985d452d2583baf19967fa76cb5c17bc9e245529d0b98745721aa7200ecaf7a",
+  "pnpm": {
+    "overrides": {
+      "@expo/dom-webview": "56.0.1"
+    }
+  },
```
andrascodes: I'm a bit paranoid about making monorepo-wide overrides. I think then we always need to keep it updated or if we start doing it and we keep extending the list then we might end up with an obscure bug that is the result of one of these overrides that caused a version mismatch either in one of the apps (in the good case) or in the packages which we might only discover after publishing and trying out in a real app.

For this `dom-webview`: Claude tells me that it's being pulled in by `@expo/log-box@56.0.14` which is probably a bug with that dependency. Since the `dom-webview` is only used in `expo-example` we can just install it as an explicit dep of the `expo-example` app itself. That was my solution [here](https://github.com/zerodevapp/zerodev-wallet-sdk/pull/250/changes#diff-cdf4b9abebf86a912a4da98c8cd90f6d0facea77f52e968e54d777779d5c3f91R14). What do you think?
I think it's better if the `expo-example` app's `package.json` or the `metro.config` becomes a  mess instead of the whole monorepo's `package.json`.

andrascodes (changes requested):
Only requesting changes for the override in the monorepo's `package.json`. Otherwise all good and thanks so much for figuring this out and testing it!!! 💙

andrascodes (approved):
Awesome!! Thanks so much!!!

PR #291, opened 2026-07-03
feat(react-ui): add size variants (sm/md/lg) via a density-scaled spacing system

brtkx:
## Summary

Adds `sm`/`md`/`lg` size variants to the wallet UI. A single `--zd-density` variable proportionally scales the whole `AuthFlow`/`Screen` card — spacing, control heights, frame dimensions, and type — exposed via a `size` prop.

## Examples from other libraries

Other libraries scale a whole UI from one variable the same way:
   - Tailwind v4 --spacing (the mechanism this builds on):
     https://tailwindcss.com/docs/margin
   - Radix Themes `scaling` prop (90%–110%) via a --scaling var:
     https://www.radix-ui.com/themes/docs/theme/spacing
   - shadcn/ui --radius (calc(var(--radius) * N)):
     https://ui.shadcn.com/docs/theming
   - Material UI theme.spacing factor (build-time variant):
     https://mui.com/material-ui/customization/spacing/

## Changes

**Density system** (`packages/wallet-react-ui/src/styles.css`)
- `--zd-density` (unitless, default 1) drives `--zd-spacing`; scoped per `[data-zd-size]` so each Screen instance recomputes locally and cascades to its subtree. Maps `sm`→0.8, `md`→0.9, `lg`→1.
- Reference links to equivalent single-variable scaling in Tailwind v4, Radix, shadcn, MUI.

**Token wiring** (`packages/react-ui/tailwind.config.ts`)
- Custom spacing steps (`13`/`4.5`, new `17`) now derive from `--zd-spacing` instead of literal px, so `h-13`/`h-17`/etc. scale.
- Font-size tokens multiply by `--zd-density`, so text scales with the variants.

**`size` prop** (`Screen`, `auth/AuthFlow`)
- New `size?: 'sm' | 'md' | 'lg'` (default `lg`), written as `data-zd-size` on the Screen root.
- Screen frame: `w-100`/`h-202.5` (400×810 at density 1) now density-scaled; `rounded-[38px]` outer radius = inner 32px + 6px ring so the gradient border has uniform width.
- `TopNav` height and content top-padding switched from fixed px to `--zd-spacing` multiples so they scale with the frame (fixed px would overflow at smaller sizes).

**Component sizing** (`react-ui`)
- `Input`/`ListItem` row + icon boxes moved from arbitrary px (`h-[68px]`, `w-[52px]`) to spacing steps (`h-17`, `h-13`/`w-13`).
- Input text bumped to `text-body1` (was unset → native fallback); `PoweredBy` logo `h-[18px]`→`h-4.5`.

**Demo** (`apps/zerodev-signer-demo`)
- Landing renders `<AuthFlow size="md" />` directly; drops the manual scale/fixed-width wrapper and widens the grid column 360→400px.
- `next.config.ts`: add `@zerodev/wallet-react-ui` to `transpilePackages` (single wagmi instance in the bundle).

## Breaking changes

- `Screen` default is `size="lg"`. Bare `<Screen>`/`<AuthFlow>` render at density 1 (unchanged visual), but the pre-existing mobile full-screen behavior (`fixed inset-0 h-[100dvh]`) is removed — Screen is now a parent-relative box at all breakpoints (`max-w/h-full`). Consumers relying on viewport-pinning must wrap it.

## Notes

- Corner radii, hairline borders, and icon sizes are intentionally fixed (don't scale with density).
- Arbitrary-value sizes in signing components (`h-[84px]`, etc.) are outside the auth surface and untouched — they won't scale.

### DIFF
--- a/apps/zerodev-signer-demo/src/app/page.tsx
+++ b/apps/zerodev-signer-demo/src/app/page.tsx
@@ -84,7 +84,7 @@ function LandingPageInner() {
     <div className="min-h-screen">
       <AppHeader/>
       <main
-        className="mx-auto grid min-h-[calc(100vh-88px)] w-full max-w-[1040px] items-center gap-x-10 gap-y-4 px-4 py-8 sm:px-6 lg:grid-cols-[1fr_360px] lg:px-0 lg:py-12">
+        className="mx-auto grid min-h-[calc(100vh-88px)] w-full max-w-[1080px] items-center gap-x-10 gap-y-4 px-4 py-8 sm:px-6 lg:grid-cols-[1fr_400px] lg:px-0 lg:py-12">
         <section className="mx-auto max-w-xl text-center lg:mx-0 lg:text-left">
           <h1 className="font-[var(--font-dm-sans)] text-4xl font-bold leading-[1.06] text-[var(--ink)] sm:text-5xl">
             Embedded Smart Wallets
@@ -112,7 +112,7 @@ function LandingPageInner() {
           </div>
         </section>
 
-        <div className="mx-auto flex w-full max-w-[360px] flex-col items-center lg:mx-0">
+        <div className="mx-auto flex w-full flex-col items-center lg:mx-0">
           {showReconnect ? (
             <div className="flex h-[729px] w-[360px] items-center justify-center">
               <button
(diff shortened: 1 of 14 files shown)


apps/zerodev-signer-demo/src/app/page.tsx
```diff
@@ -124,13 +124,9 @@ function LandingPageInner() {
...
           ) : (
-            <div className="h-[729px] w-[360px] overflow-hidden">
-              <div className="origin-top-left scale-[0.9]">
-                <div className="w-[400px]">
-                  <AuthFlow/>
-                </div>
-              </div>
-            </div>
+
+              <AuthFlow size="md" />
```
OmarBasem, line 128: I think it should be consistent with the `AuthFlow` component in `verify/page.tsx`?

brtkx: I've added that to verify too, however we won't really see a difference because we use OTP to verify.

OmarBasem (approved):

PR #292, opened 2026-07-03
fix(wallet-react): support wagmi v3 and move ox to dependencies

andrascodes:
Chris found an issue when following the documentation where wagmi v3 raised peer-dependency conflict errors. The below details all the problems and the solutions. There's one somewhat unresolved issue with `@zerodev/sdk` using `events` that is not shipped with RN. I suggested a couple of solutions, let me know what you prefer!

## What

`packages/react/package.json` (`@zerodev/wallet-react`):

- `wagmi` peer: `^2.19.0` → `^2.19.0 || ^3.0.0`
- `@wagmi/core` peer: `^2.22.0` → `^2.22.0 || ^3.0.0`
- `ox`: moved from `peerDependencies` (`^0.3.0`) to `dependencies` (`~0.14.20`)

No code changes. `@zerodev/wallet-core` unchanged.

## Why

- **wagmi v3**: a plain `npm i wagmi` now installs v3, which the v2-only peer
  rejected (`ERESOLVE`). Our wagmi usage (`useConfig`, `useChainId`,
  `useReadContract(s)`, `createConnector`, the `@wagmi/core/actions` `connect`,
  `Config.emitter`/`transports`) is unchanged across v2→v3, so widening is safe.
  They stay **peers** because wagmi must be a single shared instance — the
  connector plugs into the app's `Config`. Widening also lets npm dedupe to one
  `@wagmi/core` (the v2-only peer caused two copies, so our connector could end
  up registered in a different `@wagmi/core` than the app's `WagmiProvider`).
- **ox**: we only use `Provider.createEmitter` (a framework-agnostic EIP-1193
  emitter — no shared identity needed), and `ox` ships breaking changes across
  `0.x` minors. As a hard peer it forced `^0.3.0` to reconcile against the rest
  of the stack (viem/accounts on `0.14.x`) and failed. viem and `accounts`
  already treat `ox` as a private `dependency`; we now match that. `~0.14.20`
  dedupes onto viem's pinned `0.14.29`.

## React Native `events` error

On Expo SDK 57, bundling fails:

```
The package at "node_modules/@zerodev/sdk/.../KernelEIP1193Provider.js"
attempted to import the Node standard library module "events".
It failed because the native React runtime does not include the Node standard library.
```

`@zerodev/sdk` imports Node's built-in `events`, which React Native doesn't ship.
It surfaces with `npm` because npm doesn't auto-install the (optional) peer-dep
chain that would otherwise pull `events` in transitively (e.g. WalletConnect) —
so nothing provides it, and the import fails. Package managers that do install
those peers (e.g. pnpm in the reference app) happen to have `events` and never
hit this.

**Quick fix (needs documenting):** consumers install the pure-JS `events` shim
alongside the other packages:

```bash
npm i events
```

Update the Expo quickstart to include this, alongside the existing
`react-native-get-random-values` polyfill.

**More durable fix (up for discussion)** — so consumers don't add the shim
manually:

1. Add `events` as a regular **dependency** of `@zerodev/wallet-react`, so every
   consumer gets it automatically (same approach as `ox` above). Smallest, in
   our control.
2. Fix upstream in `@zerodev/sdk` — add a `react-native`/`browser` export
   condition, or swap Node `events` for a cross-platform emitter (`eventemitter3`,
   or `ox`'s `Provider.createEmitter` we already use). Cleanest long-term, but a
   different package.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -86,14 +86,13 @@
   "license": "MIT",
   "peerDependencies": {
     "@tanstack/react-query": "^5.0.0",
-    "@wagmi/core": "^2.22.0",
+    "@wagmi/core": "^2.22.0 || ^3.0.0",
     "@zerodev/ecdsa-validator": "^5.4.9",
     "@zerodev/sdk": "5.5.4",
     "@zerodev/wallet-core": "workspace:*",
-    "ox": "^0.3.0",
     "react": "^18.0.0 || ^19.0.0",
     "viem": "^2.38.0",
-    "wagmi": "^2.19.0",
+    "wagmi": "^2.19.0 || ^3.0.0",
     "zustand": "^5.0.3",
     "expo-linking": ">=7.0.0",
     "expo-web-browser": ">=14.0.0",
@@ -117,6 +116,9 @@
     "uuid": {
       "optional": true
     }
+  },
+  "dependencies": {
+    "ox": "~0.14.20"
   },
   "devDependencies": {
     "@types/react": "^19",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -49,13 +49,13 @@ importers:
     dependencies:
       '@expo/dom-webview':
         specifier: 56.0.1
-        version: 56.0.1(expo@56.0.13)(react-native@0.85.3(@babel/core@7.29.0)(@react-native/metro-config@0.85.0(@babel/core@7.29.0)(bufferutil@4.0.9)(utf-8-validate@6.0.6))(@types/react@19.2.14)(bufferutil@4.0.9)(react@19.2.3)(utf-8-validate@6.0.6))(react@19.2.3)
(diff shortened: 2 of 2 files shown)


brtkx (approved):

SahilVasava (commented):
Code changes LGTM — wagmi v3 peers + `ox → dependencies` are the right fixes.

**On the open `events` question — document it, don't add it to `wallet-react`.** On RN the consumer already sets up Node polyfills (`events`, `stream`, `crypto`, `buffer`, random-values), and `@zerodev/sdk`'s RN guidance already covers that — `events` is just one entry in that list. Adding it to `wallet-react` deps only removes one of several polyfills the app still has to configure, and introduces an odd direct dependency on a Node builtin that's only missing on RN (web/Node have it natively). So: list `events` in the RN polyfill setup (next to the others / mirroring `@zerodev/sdk`'s RN docs) — no dep change here.

**Minor (non-blocking):** `ox: ~0.14.20` patch-pins to viem's current ox; if viem bumps ox to `0.15` you'll get two ox copies. A range that tracks viem's ox avoids that, but it's fine for now.

**Before merge:** add a **changeset** (patch on `@zerodev/wallet-react`) so it flows into a release.

_(The red CI is the shared email-deliverability e2e flake, not this PR — Build/Type Check/Lint/Test are green.)_

PR #299, opened 2026-07-07
feat(sra): TokenChainPill component

OmarBasem:
Closes [FS-2384](https://linear.app/offchain-labs/issue/FS-2384/tokenchainpill-component)

### Summary

This PR implements TokenChainPill component

### DIFF
--- a/packages/smart-routing-address-react-ui/src/components/TokenChainPill/index.tsx
+++ b/packages/smart-routing-address-react-ui/src/components/TokenChainPill/index.tsx
@@ -1,0 +1,108 @@
+import { cn, Icon, Text, Wrapper } from '@zerodev/react-ui'
+import type { KeyboardEvent } from 'react'
+
+export interface TokenChainPillProps {
+  /** Text label rendered next to the logo (e.g., "USDC", "Base"). */
+  label: string
+  /** URL of the logo image; when omitted, a `logoBg` + `logoInitial` placeholder is drawn. */
+  logoUri?: string
+  /** Fallback background color when no `logoUri` is supplied. */
+  logoBg?: string
+  /** Fallback initial letter shown inside the placeholder circle. */
+  logoInitial?: string
+  /** Click handler; when supplied and not `disabled`, the pill becomes a keyboard-accessible button. */
+  onClick?: () => void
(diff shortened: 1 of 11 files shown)


packages/smart-routing-address-react-ui/src/styles.css
brtkx, line 1: Not part of this PR but once we have UI ready we should also remember to add the same [density system](https://github.com/zerodevapp/zerodev-wallet-sdk/blob/main/packages/wallet-react-ui/src/styles.css) we implemented for the wallet.

packages/smart-routing-address-react-ui/src/components/TokenChainPill/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  /** Fallback initial letter shown inside the placeholder circle. */
+  logoInitial?: string
+  /** Click handler; when supplied and not `disabled`, the pill becomes a keyboard-accessible button. */
+  onClick?: () => void
+  /** When true, renders as a dimmed, non-interactive pill (no chevron). */
+  disabled?: boolean
+  className?: string
+}
+
+export function TokenChainPill({
```
brtkx, line 20: Isn't this basically a `Select` primitive with an optional image? Based on its looks and function, it seems like it should be part of `react-ui`?

brtkx: If we did that, we could have [this component](https://www.figma.com/design/gn9boReFlpZSOBZkoF0LiX/ZeroDev-Wallet-UX?node-id=17777-81278&m=dev) as `Select` as well. It just disabled and has chevron hidden.

OmarBasem: That component is already included

OmarBasem: > Isn't this basically a Select primitive with an optional image? Based on its looks and function, it seems like it should be part of react-ui?

Currently it is being used in SRA only. If we needed somewhere else, it can then be refactored to `react-ui`

brtkx (approved):

PR #302, opened 2026-07-07
feat(react-ui): replace Icon zerodevLogo with a dedicated ZeroDevLogo component

brtkx:


packages/wallet-react-ui/src/shared/components/PoweredBy/PoweredBy.test.tsx
brtkx, line 1: Circular tests and tests asserting for styling. Removed.

OmarBasem (approved):

PR #303, opened 2026-07-07
fix(e2e): switch temp-email provider from mail.tm to Guerrilla Mail

brtkx:


e2e/helpers/temp-email-mailtm.ts
brtkx, line 1: Only renamed previous `temp-email.ts` so that we can use it if needed

christopherferreira9 (approved):

andrascodes (approved):

PR #304, opened 2026-07-08
feat(sra): LabeledValueRow component

OmarBasem:
Closes [FS-2389](https://linear.app/offchain-labs/issue/FS-2389/labeledvaluerow-component)


packages/react-ui/src/components/DataRow/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+            data-testid="labeled-value-row-info"
+          >
+            <Icon
+              name="info"
+              className="zd:size-3.5 zd:text-greyScale/50"
+              aria-hidden
+            />
+          </button>
+        ) : (
+          <Icon
```
brtkx, line 86: This case only renders info icon user can't interact with; it doesn't serve any purpose. IMO we should do two things here:
1. Move [DataRow](https://github.com/zerodevapp/zerodev-wallet-sdk/tree/main/packages/wallet-react-ui/src/signing/components/DataRow) to `react-ui` and reuse it here.
2. Add `labelTooltip: string` optional property and render info icon with tooltip for this case. The consumer shouldn't need to pass the tooltip behavior, unless I'm missing something.

OmarBasem: @brtkx refactored to a single component in react-ui

OmarBasem:
@brtkx added changeset file

andrascodes (approved):

PR #306, opened 2026-07-08
refactor(react-ui): ArrowCardPair

OmarBasem:
### Summary

This PR refactor `ArrowCardPair` component to `react-ui` as it is reused in SRA.


packages/react-ui/src/components/ArrowCardPair/index.tsx
brtkx, line 1: I never really liked the `ArrowCardPair` name, feels very verbose. How about `ArrowStack` or `CardPair`?

packages/react-ui/src/index.ts
brtkx, line 1: `react-ui` needs changeset update since now we have automatic releases. It will be a patch change. `wallet-react-ui` doesn't need it because nothing changes on the consumer end.

[Here's](https://github.com/zerodevapp/zerodev-wallet-sdk?tab=contributing-ov-file#the-flow--two-merges-zero-manual-publishing) the full reference on how to do it. One note: we don't need to merge https://github.com/zerodevapp/zerodev-wallet-sdk/pull/316 - we can just keep on developing until we have a decent stack of changes or need an urgent release.

packages/react-ui/src/components/ArrowCardPair/index.tsx
OmarBasem: Honestly I prefer `ArrowCardPair` over the other 2, but I can consider renaming it to something else. I was thinking maybe `CardPairWithArrow` although that's even longer, but more understandable I think.

Anyways I would prefer to rename it in a separate PR as that component touches several files that would cause many conflicts.

brtkx: Let's leave it as is then, it's more of a nit.

brtkx (approved):

PR #315, opened 2026-07-08
feat: SRA first page UI

OmarBasem:
Closes [FS-2394](https://linear.app/offchain-labs/issue/FS-2394/sra-first-page)

### Summary

This PR implements SRA first page in the designs.

Added new SRA demo app.


packages/react-ui/src/index.ts
brtkx, line 1: Needs changeset

packages/react-ui/src/index.ts
```diff
@@ -30,8 +30,11 @@ export {
   type ListItemProps,
   ListItemSkeleton,
 } from './components/ListItem'
+export { PoweredBy } from './components/PoweredBy'
+export { Screen } from './components/Screen'
 export { Switch, type SwitchProps } from './components/Switch'
 export { Text, type TextProps } from './components/Text'
+export { TOP_NAV_HEIGHT, TopNav } from './components/TopNav'
```
brtkx: I don't think `TOP_NAV_HEIGHT` needs to be exported? Ideally we shouldn't need to export a constant value like this from UI library.

packages/react-ui/src/components/TopNav/index.tsx
```diff
@@ -29,6 +33,8 @@ export function TopNav({
     >
       {onBack ? (
```
brtkx: This is not going to scale very well, and imo we should address it right away - especially because it's a `react-ui` component now shared across projects. What we could do is the following.
1. Change TopNav props to:
```
onBack -> onLeftButtonClick
onClose -> onRightButtonClick
leftButtonIcon = "chevronLeft"
rightButtonIcon = "x"
```
2. Adjust the component to:
```typescript
onLeftButtonClick && <IconButton iconName={leftButtonIcon} onClick={onLeftButtonClick} />
onRightButtonClick && <IconButton iconName={rightButtonIcon} onClick={onRightButtonClick} />
```

This way we have all we need to craft the help button + we never have to expand react-ui API in the future.

brtkx: Technically it's a breaking change so that'd be a major version bump, but because we're at pre-1.0.0 we can just set changeset to patch.

apps/sra-demo/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@zerodev/sra-demo",
+  "version": "0.1.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "dev": "next dev -p 8004",
+    "hot-dev": "trap 'kill 0' SIGINT; pnpm --filter @zerodev/react-ui exec vite build --watch & pnpm --filter @zerodev/smart-routing-address-react-ui exec vite build --watch & (sleep 2 && open http://localhost:8004) & pnpm --filter @zerodev/sra-demo dev",
```
andrascodes, line 8: what's this `hot-dev` command exactly? It rebuilds the 2 packages and restarts the server?

packages/smart-routing-address-react-ui/src/pages/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+ */
+export interface SmartRoutingAddressProps {
+  /** Recipient address the smart routing address is created for. */
+  recipient: Address
+  /** Called when the top-right × close button is clicked. */
+  onClose: () => void
+  /** Called when the top-left ? help button is clicked. Optional. */
+  onHelp?: () => void
+  /** Called when the QR icon inside `AddressDisplay` is clicked. Optional. */
+  onQrClick?: () => void
```
brtkx, line 47: What is the purpose (or future intention) of clicking the QR code?

packages/react-ui/src/components/ArrowCardPair/index.tsx
```diff
@@ -116,11 +116,50 @@ export interface ArrowCardPairProps {
 }
 
 export function ArrowCardPair({ topCard, bottomCard }: ArrowCardPairProps) {
+  const topRef = useRef<HTMLDivElement>(null)
+  const [topHeight, setTopHeight] = useState<number>()
+
+  // Track the top card's rendered height so the arrow lands at the junction
+  // between the two cards, not the vertical center of the container. Centering
+  // in the container drifts when the two cards have unequal heights (e.g. the
+  // bottom card has extra rows), pushing the arrow off the seam.
```
andrascodes: Maybe I'm totally wrong on this. But couldn't we achieve this with some CSS? Like `position: absolute;` (with the container having `position: relative` and then manipulating the `top:` value?
So basically you just take the arrow out of the document flow and position it relative to either the the entire `Top+Bottom card` container `div` or have a `div` with zero width/height in between the 2 cards...etc.
This just feels a bit like it's an overkill for this. But again I'm definitely not all clear on the reqs and the possibilities.

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,250 @@
...
+    <div className="zd:flex zd:w-full zd:items-start zd:gap-1">
+      <div style={{ width: 162, flexShrink: 0 }}>{left}</div>
+      <div className="zd:min-w-px zd:flex-1">{right}</div>
+    </div>
+  )
+}
+
+/** Renders the address slot inside the "Arrives as" card based on the
+ * address-generation state. */
+function renderAddressSlot({
```
andrascodes: I feel like this could just be a component as well.

packages/smart-routing-address-react-ui/src/pages/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * this pass; QR / active-deposits / history / error screens will land as
+ * additional cases once their Figma is wired.
+ */
+export type SmartRoutingAddressStep = 'deposit'
+
+const TITLE_BY_STEP: Record<SmartRoutingAddressStep, string> = {
+  deposit: 'Deposit funds',
+}
+
+function renderStep(
```
andrascodes, line 18: and this could be just inlined as a ternary?

andrascodes (approved):
No big issues that block a release. A few nits and an alternative idea for the ArrowCardPair but not sure if it actually solves the issue.

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const estimatedFee = feeData
+    ? `${formatDisplayAmount(feeData.fee, feeData.decimal, 'up')} ${sourceSymbol}`
+    : '—'
+
+  const minDepositAmount =
+    feeData && sourceSymbol
+      ? `${formatDisplayAmount(feeData.minDeposit, feeData.decimal, 'up')} ${sourceSymbol}`
+      : null
+
+  const sourceChainName = source?.chain.name ?? destChain.name
```
brtkx: Why does `sourceChainName` default to destination? Is that the product requirement we have confirmed?

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,250 @@
...
+              variant="ghost"
+              className="zd:relative zd:flex zd:w-full zd:flex-col zd:rounded-2xl zd:p-1"
+            >
+              <CardTitle>Send</CardTitle>
+              <PillRow
+                left={
+                  <TokenChainPill
+                    label={sourceSymbol ?? '—'}
+                    logoBg="#2775CA"
+                    logoInitial={(sourceSymbol ?? '?').charAt(0)}
```
brtkx: We don't need to pass `logoInitial`, `TokenChainPill` can infer it from `label`

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+  const minDepositAmount =
+    feeData && sourceSymbol
+      ? `${formatDisplayAmount(feeData.minDeposit, feeData.decimal, 'up')} ${sourceSymbol}`
+      : null
+
+  const sourceChainName = source?.chain.name ?? destChain.name
+
+  // formatDuration returns "~3 min" — Figma spec is "≈ 3 min".
+  const readyInText = fillTime.replace('~', '≈ ')
```
brtkx: We shouldn't rely on `fillTime` never introducing a breaking change for us here. We can just accept `fillTime` as is. The design content is just placeholder values.

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,250 @@
...
+          text={`Watching for your deposit on ${sourceChainName}…`}
+        />
+      </div>
+
+      <PoweredBy className="zd:justify-center" />
+    </div>
+  )
+}
+
+/** Card title text (18px, VC Cardinal Wide feel). Extracted to keep the
```
brtkx: We shouldn't be hardcoding px values outside tailwind if we can avoid it. It doesn't need to be pixel perfect with the designs, lets just round to the closest tw values.

packages/smart-routing-address-react-ui/src/pages/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /** Recipient address the smart routing address is created for. */
+  recipient: Address
+  /** Called when the top-right × close button is clicked. */
+  onClose: () => void
+  /** Called when the top-left ? help button is clicked. Optional. */
+  onHelp?: () => void
+  /** Called when the QR icon inside `AddressDisplay` is clicked. Optional. */
+  onQrClick?: () => void
+  className?: string
+  size?: 'sm' | 'md' | 'lg' | undefined
```
brtkx: Don't think we need `undefined`? It's already optional.

brtkx (changes requested):

packages/react-ui/src/index.ts
andrascodes: you can always add that later though. We don't necessarily need to make a changeset per PR. We could just make a changeset PR later for multiple changes that will make a release.
Or did we agree on a convention that we include a changeset in every PR?

brtkx: If we merge https://github.com/zerodevapp/zerodev-wallet-sdk/pull/316 before we merge the changeset later, it won't be included with the specific release. Also not sure if we will remember later, we'd need to go through previous features and add everything retroactively.

apps/sra-demo/package.json
OmarBasem: `hot-dev` command is for auto-refresh to the demo app when making code updates to any of the packages, otherwise we would have to rebuild the packages and restart the app server

packages/react-ui/src/components/ArrowCardPair/index.tsx
OmarBasem: You are right that was an overkill, fixed it

packages/smart-routing-address-react-ui/src/pages/index.tsx
OmarBasem: That switch case is used to decide which page to render, currently there is one page only, but more pages will be added

packages/smart-routing-address-react-ui/src/pages/index.tsx
OmarBasem: temporary placeholder, removed [here](https://github.com/zerodevapp/zerodev-wallet-sdk/pull/322/changes#diff-b1704438fd24b6284df385a6950c940c8509929df6db86d745f7045dafc99648L46-L47)

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
OmarBasem: that was a defensive fallback, but should be removed

OmarBasem:
@brtkx addressed your comments

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,223 @@
...
+  useNewDeposits(deposits, hasLoaded)
+
+  const { destChain, destSymbol, sourceSymbol, feeData, fillTime } =
+    useMemo(() => {
+      const destChain = resolveDestChain(config)
+      const destSymbol = getDestTokenSymbol(config, source)
+      return {
+        destChain,
+        destSymbol,
+        sourceSymbol: source ? getSourceTokenSymbol(source) : destSymbol,
```
brtkx: This is also defaulting to `destSymbol` but I guess we will remove it once picker is wired?

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,223 @@
...
+export function Deposit({ onQrClick }: DepositProps) {
+  const { config, addressState } = useSmartRoutingAddressContext()
+
+  const success = addressState.status === 'success' ? addressState : null
+  const address = success?.address
+  const estimatedFees = success?.estimatedFees ?? []
+
+  // allowPartialRoutes lets the server drop source tokens it can't route, so
+  // the routable set is exactly the tokens the fee estimates came back with.
+  const srcTokens = useMemo(
```
brtkx: tbh both useMemos don't really do much except for bloating the code. `success?.estimatedFees ?? []` is introducing some code smell too because on loading it creates a new array reference, meaning it re-triggers memoized functions anyway.

I'd just inline everything instead - the cost of running these memos is probably comparable to not memoizing at all.

```typescript
const srcTokens = sourceTokensFromFees(estimatedFees)
const source = srcTokens[0] ?? null

const destChain = resolveDestChain(config)
const destSymbol = getDestTokenSymbol(config, source)
const sourceSymbol = source ? getSourceTokenSymbol(source) : destSymbol
const feeData = source
  ? findFeeData(estimatedFees, source.chain.id, source.tokenType)
  : null
const fillTime = formatDuration(
  resolveFillTimeSeconds(config, source?.chain.id ?? destChain.id),
)
```

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -0,0 +1,223 @@
...
+function PillRow({
+  left,
+  right,
+}: {
+  left: React.ReactNode
+  right: React.ReactNode
+}) {
+  return (
+    <div className="zd:flex zd:w-full zd:items-start zd:gap-1">
+      <div style={{ width: 162, flexShrink: 0 }}>{left}</div>
```
brtkx: Custom width won't scale with our density system

packages/react-ui/src/components/ArrowCardPair/ArrowCardPair.stories.tsx
```diff
@@ -31,12 +31,12 @@ type Story = StoryObj<typeof meta>
 export const Default: Story = {
   args: {
     topCard: (
-      <Wrapper variant="ghost" className="zd:rounded-2xl zd:p-4">
+      <Wrapper variant="ghost" className="zd:rounded-2xl zd:p-4 zd:bg-red-600">
```
brtkx: why red bg?

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
OmarBasem: Yes, defaulted to undefined for now

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
OmarBasem: I agree, refactored

packages/react-ui/src/components/ArrowCardPair/ArrowCardPair.stories.tsx
OmarBasem: removed

packages/smart-routing-address-react-ui/src/components/TokenChainPill/index.tsx
```diff
@@ -21,7 +24,7 @@ export function TokenChainPill({
   label,
   logoUri,
   logoBg = '#E6EFFB',
-  logoInitial,
+  logoInitial = label.charAt(0).toUpperCase(),
```
brtkx, line 27: nit: approving the PR but I don't think `logoInitial` should even be exposed. We can just use `label.charAt(0).toUpperCase()` in the UI directly? But ff to merge if it's needed. Just pointing out.

brtkx (approved):

PR #318, opened 2026-07-08
test(tests-bypass-email): Use mocked email on testing 

christopherferreira9:
## Summary 
Previously, all E2E and integration tests required hitting Guerrilla Mail and Turnkey to complete OTP flows, making them brittle and slow due to external rate limits. This PR introduces a USE_REAL_EMAIL flag that switches tests between real and mocked backends — PR runs use mocked mode (no external dependencies), and a new daily scheduled workflow runs with real email to catch live-service regressions.

### SDK side changes
---
`packages/core/src/core/createZeroDevWalletCore.ts`

Added dangerouslyOverrideOtpSignerPublicKey?: string to ZeroDevWalletConfigCore. When present, it's forwarded to encryptOtpAttempt as dangerouslyOverrideSignerPublicKey, bypassing the production-pinned TLS Fetcher key check. The spread conditional (...(config.x && { ... })) ensures the property is only passed when explicitly set — it never appears in a normal integrator's call.

`packages/react/src/core/connector.ts`

Same option added to ConnectorCoreParams. It's forwarded into createZeroDevWallet() using the same conditional spread — so an integrator who doesn't set it gets no property at all, not undefined.

Both additions are marked @internal with a "Never set this in production" JSDoc comment.

---
**Tests added**

`packages/core/src/utils/encryptOtpAttempt.test.ts` — new test: generates a fresh ephemeral key pair (not the production pinned key), builds a test bundle signed with it, and asserts that calling encryptOtpAttempt without the override throws does not match pinned signing key. This proves the default path enforces the production key.

`packages/react/src/connector.test.ts` — two new tests in a dangerouslyOverrideOtpSignerPublicKey security guard describe block:
- A connector created without the option must not forward it to createZeroDevWallet — guards against accidental propagation.
- A connector explicitly configured with it must forward it correctly — proves the intentional path works.

### Workflow changes
---
#### `ci.yml`
1. Renamed test job from "Test" → "Unit Tests" 
2. Extracted integration and E2E jobs into a reusable workflow (test.yml). The two inline jobs (integration-test, e2e-test) were removed from ci.yml and replaced with a single call-tests job that delegates to test.yml. This also drops the dependency on the `build` job for parallel execution in favor of faster feedback from tests.
3. Added all-checks-pass gate job - a job that acts as a single source of truth for what jobs require a passing status in order for the merge to be possible. This reduces management on the repository settings as we can remove and add _mandatory_ jobs through code. The repository then only needs to have `all-checks-pass` passing.

#### `test.yml` - new workflow
Reusable workflow extracted from `ci.yml`. It contains:
-  **Integration Tests** — runs `pnpm test:integration`
-  **Browser E2E Tests** — builds the SDK, starts the demo app, runs Playwright
Both of these workflows accept a newly added input called `use_real_email` with the default value being `true`. `ci.yml` then sends `use_real_email` as `false` in order to remove the dependency on the email service on PRs and main runs (we may consider using the real email service for main runs - to be discussed).

#### `test.yml` - new workflow
Runs test.yml daily at 08:00 UTC with use_real_email: true (real Guerrilla Mail + Turnkey). Also triggerable manually via workflow_dispatch with the option to toggle email mode.

### Concerns
---
**1. Mock fidelity**
As the payload returned by the backend can change, fixtures can become stale and results would not be reliable. A `fixture-shape.test.ts` test has been created to catch this drift on real email mode. On a future development, the PR t update the fixture can be made automatically on failure.

**2. The rejection test tests the mock, not the logic**
"should reject an invalid OTP code" in mock mode stacks setupNodeMocks({ rejectOtpVerify: true }), which unconditionally returns HTTP 400 regardless of what was sent. It doesn't verify that a semantically wrong OTP (wrong code, wrong encryption) is actually detected — it verifies that an HTTP 400 throws an exception. This test is still valued as we have UI checks - we should rename this test if we intend to keep it.

**3. dangerouslyOverrideSignerPublicKey is a scattered pattern**
Every integration test call site that invokes encryptOtpAttempt directly must manually pass dangerouslyOverrideSignerPublicKey: isRealEmail() ? undefined : MOCK_OTP_SIGNER_PUBLIC_KEY. This is currently in three places. Anyone writing a new integration test that calls encryptOtpAttempt directly will get a cryptic key-mismatch error in mock mode until they discover the pattern. This can be mitigated.

**4. Fork PRs have no integration test coverage at all**
call-tests is skipped on fork PRs (no secret access). all-checks-pass treats skipped as OK, so external contributors get zero integration or E2E feedback. A fork PR that breaks the OTP flow in a way that unit tests don't catch can pass all required checks. I personally think this is acceptable.

**5. otp-init-shape.test.ts is permanently skipped in mock mode** 
This test exists specifically to validate the shape of the real backend's OTP init response and verify the production pinned key. Skipping it in mock mode means the CI run that runs on every PR never exercises it. It only runs in the daily scheduled run. If the production key rotates or the bundle shape changes, a PR that breaks the integration won't be caught at PR time.

**6. Passkeys can't be tested when the mock mode is enabled**


.github/workflows/ci.yml
```diff
@@ -92,7 +92,7 @@ jobs:
           done
 
   test:
-    name: Test
+    name: Unit Tests
```
christopherferreira9, line 95: Renamed for a clear read on the pipeline since it hasn't been extracted to the dedicated test workflow.

.github/workflows/ci.yml
```diff
@@ -113,93 +113,30 @@ jobs:
...
-          ZD_PROJECT_ID: ${{ secrets.ZD_PROJECT_ID }}
-
-  e2e-test:
-    name: Browser E2E Tests
+    uses: ./.github/workflows/test.yml
+    with:
+      use_real_email: false
+    secrets: inherit
+
+  all-checks-pass:
```
christopherferreira9, line 124: single job to have a mandatory success status. Easier to maintain through code.

.github/workflows/scheduled-tests.yml
```diff
@@ -0,0 +1,20 @@
+name: Scheduled Tests
+
+on:
+  schedule:
+    - cron: '0 8 * * *'
```
christopherferreira9, line 5: Currently scheduled to run daily using using real email keeping the original behavior and reducing the amount of flakiness due to rate limiting from the email service probider.

.github/workflows/test.yml
```diff
@@ -0,0 +1,130 @@
...
+          fi
+          pnpm build
+          pnpm start &
+        env:
+          NEXT_PUBLIC_ZERODEV_PROJECT_ID: ${{ secrets.ZD_PROJECT_ID }}
+          NEXT_PUBLIC_KMS_PROXY_BASE_URL: https://kms.staging.zerodev.app/api/v1
+          # Pin the bundler/paymaster to staging so the AA stack matches the
+          # staging KMS above (the SDK default host is prod).
+          NEXT_PUBLIC_ZERODEV_AA_HOST: https://staging-meta-aa-provider.onrender.com
+          USE_REAL_EMAIL: ${{ inputs.use_real_email }}
```
christopherferreira9, line 103: Demo is built with this env variable which is then passed downstream to the wagmi options.

e2e/browser/magic-link.spec.ts
```diff
@@ -55,13 +46,16 @@ test.describe('Magic Link Flow', () => {
...
-    // Step 6: Poll for email and extract OTP code from the magic link URL
-    const emailContent = await searchForNewEmail(
-      emailAccount.authToken,
-      EMAIL_POLL_INTERVAL_MS,
-      EMAIL_POLL_TIMEOUT_MS,
-    )
-    const otpCode = extractOtpCodeFromMagicLinkUrl(emailContent)
+    // Step 6: Resolve OTP code — use fixture value in mock mode, poll inbox in real mode
+    const otpCode =
+      magicLinkSession.otpCode ??
```
christopherferreira9, line 51: when in mock mode, otpCode comes directly from the fixture. The fixture handles the real vs mocked email logic.

christopherferreira9:
This PR is on hold as with the current implementation we'll end in bypassing some operations like mint which is not ideal.

PR #319, opened 2026-07-08
chore(deps): fix audit vulnerabilities

brtkx:
Down from 180 vulnerabilities to 3 we cannot fix (2 moderate + 1 low). In the next PR I'll add audit to CI to block merges on high vulnerabilities.


apps/zerodev-signer-demo/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "incremental": true,
     "module": "esnext",
     "esModuleInterop": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
brtkx, line 13: required by one of the viem's chains after bumping version

OmarBasem (approved):

PR #321, opened 2026-07-10
feat(react-ui): SelectDropdown component

OmarBasem:
Closes [FS-2398](https://linear.app/offchain-labs/issue/FS-2398/selectdropdown-component)


packages/react-ui/src/components/SelectDropdown/index.tsx
```diff
@@ -0,0 +1,216 @@
+import {
```
andrascodes: Image

The Default story looks a bit off. It might be just an issue with the story file layout itself

packages/react-ui/src/components/SelectDropdown/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+  className,
+  panelClassName,
+  anchorRef,
+}: SelectDropdownProps) {
+  const [open, setOpen] = useState(false)
+  // `mounted` gates the portal so we don't touch `document` during SSR.
+  const [mounted, setMounted] = useState(false)
+  const [panelStyle, setPanelStyle] = useState<CSSProperties>()
+  const rootRef = useRef<HTMLDivElement>(null)
+  const panelRef = useRef<HTMLDivElement>(null)
```
andrascodes: Can't we just use a library that does all this for us (and probably more)? Like a Headless UI Dropdown? Why did you decide to implement it from zero? Maybe I'm missing something

andrascodes (changes requested):
I'm just marking it as Request for changes because of the suggestion of using a library for the SelectDropDown. If you disagree just request another review and I'll mark it approved.

packages/react-ui/src/components/SelectDropdown/index.tsx
brtkx: We had some issues with using a headless UI library before but only with expo where we had to share them between web and native. But now I agree we could use something like [Radix](https://www.radix-ui.com/primitives/docs/components/select) for all of our primitives.

I'm not very opinionated on which library we should use, if you guys prefer any other ff to suggest. I just like radix. We can start with this component and later make the rest of primitives use it.

We should also make sure we don't ship it and it's dev dep only imo.

andrascodes: I don't think we should at all think about sharing components between Expo and Web anymore. If we ever do a `wallet-react-native-ui` that will have to have a totally new component library. Hopefully, the Tailwind config we could share though.

> We should also make sure we don't ship it and it's dev dep only imo

You mean not shipping Radix? How do you do that? If we're using a `<Dropdown />` component from Radix we have to either have Radix as a full dep or peer dep. We could use subpath exports too to make sure Radix components don't make it into the bundle if we're not importing them if that's a concern.

Shadcn uses BaseUI as well alongside Radix UI so that's an option too. Or even Shadcn itself.

brtkx: I agree with your first point. My comment was that in the past we decided not to use ui lib because of expo. But now imo we should use one.

Lets discuss about specific libraries today on sync

brtkx: But I'm also fine using whatever, we can just use shadcn without having the sync. But we should have any one instead of writing our own components for sure imo.

OmarBasem:
@andrascodes @brtkx thanks for your reviews, I have refactored to using Radix

packages/react-ui/src/components/SelectDropdown/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+    <PillItem
+      label={selected.symbol}
+      {...(selected.logoUri && { logoUri: selected.logoUri })}
+      {...(selected.logoBg && { logoBg: selected.logoBg })}
+      {...(disabled && { disabled: true })}
+    />
+  )
+
+  return (
+    <Popover.Root>
```
brtkx: This is not the right component - we should be using https://www.radix-ui.com/primitives/docs/components/select

packages/react-ui/src/components/SelectDropdown/index.tsx
brtkx, line 1: I'd just call it `Select` - it's pretty standard for UI libraries.

packages/react-ui/src/components/SelectDropdown/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+  if (!selected) {
+    return (
+      <div className={cn('zd:relative', className)}>
+        <PillItem label={placeholderLabel} disabled />
+      </div>
+    )
+  }
+
+  const trigger = (
+    <PillItem
```
brtkx: I feel like `PillItem` is quite out of place in `react-ui` in general. It looks like a disabled select trigger - I would just move it back to SRA. Even if it's used in other packages later, it's just icon and text aligned horizontally. IMO this doesn't justify being in the shared package.

packages/react-ui/src/components/SelectDropdown/index.tsx
brtkx, line 1: I don't think we should be using radix this way. We should do it shadcn style where we export styled radix components and the consumer does what it wants with them. With the current implementation the code is bloated (a lot of props), and the consumer can only build token related components with it.

https://github.com/zerodevapp/zerodev-dashboard/blob/main/src/components/ui/select.tsx
ZeroDev dashboard is a good example how Select component has been exported.
Also can see here how shadcn exports radix components:
https://github.com/shadcn-ui/ui/blob/main/apps/v4/registry/bases/radix/ui/select.tsx

Feel free to take some time to dive deep into other projects, including ZeroDev dashboard, and check out how they export and consume radix.

brtkx (changes requested):

packages/react-ui/src/components/SelectDropdown/index.tsx
brtkx: https://github.com/zerodevapp/zerodev-wallet-sdk/pull/322
For example the other component you've implemented is quite nice, one thing we can do is get rid of props by utilizing radix to its full potential. Generally radix allows us to avoid using monolith components - so that the consumer can build anything with styled blocks instead of passing props.

packages/react-ui/src/components/SelectDropdown/index.tsx
OmarBasem: I think it's a very basic component (icon + text) that can be in react-ui, very likely it would be used in other packages.

packages/react-ui/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+      const selected = TOKENS.find((t) => t.id === value) ?? TOKENS[0]
+      return (
+        <Select value={value} onValueChange={setValue}>
+          <SelectTrigger asChild>
+            <PillItem label={selected.symbol} logoBg={selected.logoBg} />
+          </SelectTrigger>
+          <SelectContent align="start">
+            {TOKENS.map((t) => (
+              <SelectItem key={t.id} value={t.id}>
+                <div className="zd:px-3 zd:py-2">
```
brtkx: I'm thinking if this div wrapper should go inside the Select component? I imagine this would be the same for every consumer all the time?

packages/react-ui/src/components/Pill/index.tsx
```diff
@@ -93,7 +99,7 @@ export function TokenChainPill({
         </div>
         <Text className="zd:whitespace-nowrap zd:text-body1">{label}</Text>
       </div>
-      {interactive && (
+      {!disabled && (
         <div className="zd:flex zd:shrink-0 zd:items-center zd:rounded-full zd:p-2">
           <Icon
```
brtkx: I think chevron should just be part of the Select component and this stay an image and label? Chevron has nothing to do with this component - `Pill` should be static and `disabled` be controlled only by `Select`.

packages/react-ui/src/components/Pill/index.tsx
brtkx, line 1: I'm thinking if we should just call it a `Pill`? `Item` brings nothing to the table, it's just a redundant postfix that should really be only applicable to item of a list (e.g. SelectItem, ListItem). It doesn't change the semantic meaning of this component either, calling it `Pill` or `PillItem` hints at it being a pill component for both cases.

packages/react-ui/src/components/Select/Select.stories.tsx
brtkx: But if there was a select item that's not plain text but a react node, then it could get tricky because padding would be added to the component. wdyt?

I don't mind if we have to repeat it, I'm just a little bit concerned if we will be able to maintain the consistency. 

Does the div even need to exist? Would it work if we just added className to `SelectItem`, default to `zd:px-3 zd:py-2` and let the consumer override if needed?

packages/react-ui/src/components/Pill/index.tsx
OmarBasem: Chevron is visually part of the `Pill` component, `Select` component is not a UI component (any component can be used as the dropdown trigger) so I think chevron can stay in `Pill`. In the future chevron could be refactored to a `trailingIcon` to be more generic but currently it has one usage as chevron so I'll  keep it as it is for now

brtkx: But `Select` does export `SelectIcon` so I think we can use that. What if we refactored to the `trailingIcon` now? It should be a quick thing. Then it would look something like this, and we don't need to have a chevron in the pill. (`trailingIcon` in `Pill` can be optional, without a default)
```typescript
<SelectTrigger asChild>
  <Pill label={sourceSymbol ?? '—'} trailingIcon={<SelectIcon />} />
</SelectTrigger>
```

OmarBasem: @brtkx refactored

.changeset/silver-radix-popover.md
brtkx, line 1: Thanks for the changes, approved. One nit: changeset should only include changes we made in the PR and that are going to land in the package. 

For example, I think this one was part of addressing the comments, so we never dropped dependency from the previous version?

```
Dropped peer dependency: `@radix-ui/react-popover` — no longer used.
```

brtkx (approved):

andrascodes (approved):
Looks good! Thanks!!

PR #322, opened 2026-07-13
feat(sra): Add QR Modal

OmarBasem:
Closes [FS-2282](https://linear.app/offchain-labs/issue/FS-2282/deposit-address-qr-modal)


packages/react-ui/package.json
```diff
@@ -79,6 +80,7 @@
   },
   "dependencies": {
     "clsx": "^2.1.1",
+    "qrcode": "^1.5.4",
```
brtkx: Maybe we could use https://www.npmjs.com/package/uqr? It's more lightweight for what we're trying to do.

OmarBasem: @brtkx good suggestion, swapped

packages/smart-routing-address-react-ui/src/pages/index.tsx
```diff
@@ -70,6 +68,33 @@ export function SmartRoutingAddress({
...
+    try {
+      await navigator.clipboard.writeText(address)
+    } catch {
+      // Clipboard can reject on insecure contexts / permissions; the visual
+      // "copied" state isn't in the spec yet, so just close and swallow.
+    }
+    setQrOpen(false)
+  }
+
+  const overlay =
```
brtkx: This only works because `SmartRoutingAddress` owns `qrOpen` and renders the `Screen`. If you had state live inside a component wrapped by `Screen`, that wouldn't be possible anymore. For example:
```typescript
<Screen>
  <MyComponent />
</Screen>

// MyComponent.tsx
const [qrOpen, setQrOpen] = useState(false)

// Impossible to render the overlay
```

This is not just a speculation - it's an actual case for external wallets as you can see in https://github.com/zerodevapp/zerodev-wallet-sdk/pull/324. You can also see the implementation there.

What we could do is use a portal:
```typescript
// Screen.tsx
const ScreenOverlayContext = createContext<HTMLDivElement | null>(null)

export function useScreenOverlayTarget() {
  return useContext(ScreenOverlayContext)
}

export function Screen() {
  ...
  const [overlayTarget, setOverlayTarget] = useState<HTMLDivElement | null>(
    null,
  )

  return (
    <div ...>
      <WrapperBorder />
      <div ... ref={setOverlayTarget}>
        <ScreenOverlayContext.Provider value={overlayTarget}>
          {children}
        </ScreenOverlayContext.Provider>
      </div>
    </div>
  )
}

// SheetShell.tsx
const overlayTarget = useScreenOverlayTarget()
if (!overlayTarget) return null

return createPortal(
  <component-goes-here />,
  overlayTarget
)

// any component
<Screen>
  <SheetShell open={open} onClose={onClose}>
    ...
  </SheetShell >
</Screen>
```

Also with the current implementation we have to conditionally mount the sheet - it can be tricky to animate. With the portal it's always mounted (for as long as the component containing the sheet is rendered), easy to animate with pure css - no need for react states.

packages/react-ui/src/components/QrModal/QrCode.tsx
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * Custom QR renderer that draws runs of horizontally-adjacent modules as a
+ * single rounded pill, and the three finder patterns as rounded squares.
+ * Isolated modules naturally fall out as circles (a 1×1 pill is a circle).
+ */
+export function QrCode({
+  value,
+  size,
+  errorCorrectionLevel = 'M',
+  color = '#000',
```
brtkx: I don't think we need to expose color and bgColor. Early abstraction and QR code should probably stay black on white anyway

packages/smart-routing-address-react-ui/src/pages/index.tsx
OmarBasem: I refactored the implementation to using Radix also

packages/react-ui/src/components/SheetShell/index.tsx
```diff
@@ -0,0 +1,77 @@
+import * as Dialog from '@radix-ui/react-dialog'
```
OmarBasem: Reusable bottom sheet

packages/react-ui/src/components/SheetShell/index.tsx
brtkx, line 1: Nice looking good! A few changes I'd make here to make it even smoother:
```typescript
export const Sheet = Dialog.Root
export const SheetClose = Dialog.Close

export function SheetTitle({ className, ...props }: SheetTitleProps) {
  return <Dialog.Title className={cn('zd:sr-only', className)} {...props} />
}
```

Rename this:
```
SheetShell -> SheetContent
```

And remove this line:
```typescript
<Dialog.Root open={open} onOpenChange={onOpenChange}>
```

With this alone we can get rid of all the props, except for children and className. The consumer passes open state to `Sheet` itself. The consumer looks something like this:
```typescript
<Sheet open={isOpen}>
  <SheetContent>
    <SheetTitle>{title}</SheetTitle >
    {content}
    // If need a close button
    <SheetClose asChild>
      <Button text="Cancel" />
    </SheetClose>
  <SheetContent /> 
</Sheet>
```

So I'd drop `Shell` from the name IMO, was my initial suggestion but I think `Sheet` is enough.

OmarBasem: Refactored

packages/react-ui/src/components/BottomSheet/index.tsx
brtkx, line 1: I'd just make the name future-proof and call it a `Sheet`. So for example, in the future we can do `side: 'bottom' | 'right'` without introducing breaking changes.

Also we call it `BottomSheet` but export components like `SheetClose`, `SheetTitle`. So there's already an inconsistency we could resolve by `Sheet` only.

packages/smart-routing-address-react-ui/src/components/QrSheet/index.tsx
brtkx, line 1: I'd just keep it in the SRA package. We can't reuse it anywhere else - its content/layout is only relevant to SRA.

Maybe we will be able to do something with it in the future. But going forward when it comes to non-primitive components, even if we have duplicates for a bit, it's much better than early abstractions.

also nit: probs could name it `QrSheet` since we make a distinction between the two; and it's using `Sheet`

packages/react-ui/src/components/BottomSheet/index.tsx
OmarBasem: `BottomSheet` is not my own name, it is a common terminology used for modals that slides from the bottom: https://m3.material.io/components/bottom-sheets/overview

> in the future we can do side: 'bottom' | 'right'

In a mobile-factor design I don't think we can have that, these sheets are meant to be at the bottom

packages/smart-routing-address-react-ui/src/components/QrSheet/index.tsx
OmarBasem: Looking at the UI I don't see anything specific to SRA (except sub-title which can be a prop), I think it can be used for any deposit address not just SRA

Screenshot 2026-07-21 at 9 51 49 AM

brtkx: Yeah but we don't have any other deposit features other than SRA, and we may not have any for a while or ever; so it's an early abstraction.

brtkx: The way I see it, we can always move things to react-ui without breaking changes. Removing things from react-ui is more tricky once we are more established, because it's a breaking change and we can't just have a patch for it - it's technically a major version bump.

packages/react-ui/src/components/BottomSheet/index.tsx
brtkx: In that case all the exported components should have `BottomSheet` prefix, e.g. `SheetClose` -> `BottomSheetClose`.

packages/smart-routing-address-react-ui/src/components/QrSheet/index.tsx
OmarBasem: @brtkx refactored

.changeset/golden-radix-dialog.md
brtkx, line 1: Approved but just two last things before we merge please: 
- Don't think we need to export `useScreenOverlayContainer` from the barrel index? This seems like an internal only hook. We should only export what's used outside react-ui.
- The example code in changeset still uses old `Sheet` names.

brtkx (approved):

PR #323, opened 2026-07-13
feat(wallet-react-ui): PoC add external-wallet auth method

brtkx:
## Summary

Adds an `external-wallet` sign-in method to the kit: injected wallets via EIP-6963 auto-discovery plus an in-house WalletConnect QR step — own UI end to end, no Reown components, no runtime registry API. PoC scope.

## Done & working

- **Auth flow** — `'external-wallet'` in `AuthMethod`; SignUp gains "Continue with a wallet" → `wallet-selection` step.
- **Wallet list** — 6963-discovered + dev-configured connectors (zerodev row filtered out); click = plain wagmi `connect`, the wallet's own provider serves RPC afterwards (embedded wallet and `provider.ts` never involved).
- **Wallet guide** — `walletGuide.ts`: 10 wallets (name, `rdns`, inline SVG icon, mobile deep-link prefix, download URL). Rows currently open the vendor download page.
- **WalletConnect step** — kit creates wagmi's `walletConnect({ showQrModal: false })` from new `ZeroDevKitConfig.walletConnectProjectId` (registered via the same `_internal` path wagmi's own `connect()` uses; reuses a dev-added WC connector if present); subscribes `display_uri` *before* connecting; QR rendered in-house with `uqr`.
- **Plumbing** — `walletConnectProjectId` flows kit config → kit store (same pattern as `logo`).
- **Demo** — method enabled, PoC Reown projectId, mainnet + arbitrum added.
- **Deps** — `uqr@^0.1.3` (kit).

## Still to do (not in this PR)

- undiscovered (not installed) wallets become connect-capable

## How to test

1. `pnpm --filter @zerodev/signer-demo dev` → `localhost:3000`.
2. Injected: sign-in modal → **Continue with a wallet** → WalletConnect → Scan QR code in any wallet app → modal closes, dapp connected via the wallet's provider.


andrascodes (commented):
I like this approach a lot. Or at least I definitely can't think of a better one.
It stays more or less out of the way of our Embedded Wallet flows. It gives us the same capability that for example Privy has and it looks great. The mobile connection is also a great plus. So nice job! I think you devised a great solution!

PR #336, opened 2026-07-20
feat(sra): animate select dropdown

OmarBasem:
Closes [FS-2436](https://linear.app/offchain-labs/issue/FS-2436/animate-select-dropdown)

### Summary

This PR adds animation for Select Dropdown

### DIFF
--- a/packages/react-ui/src/components/Select/index.tsx
+++ b/packages/react-ui/src/components/Select/index.tsx
@@ -25,6 +25,12 @@ export function SelectTrigger({
       className={cn(
         'zd:inline-flex zd:items-center zd:justify-between zd:gap-2',
         'zd:outline-none zd:cursor-pointer',
+        // Radix Select sets `pointer-events: none` on the body while the
+        // panel is open (part of its scroll-lock/dismiss-layer behavior),
+        // which falls back the cursor to `default` even over the trigger.
+        // Re-enable pointer events + cursor when the trigger is open so
+        // the hover cue stays consistent.
+        'zd:data-[state=open]:pointer-events-auto zd:data-[state=open]:cursor-pointer',
         'zd:disabled:cursor-not-allowed zd:disabled:opacity-50',
         className,
       )}
@@ -76,15 +82,30 @@ export function SelectContent({
         ref={ref}
         position={position}
         sideOffset={sideOffset}
-        className={cn('zd:z-50 zd:max-h-80 zd:outline-none', className)}
+        className={cn(
+          'zd:z-50 zd:max-h-80 zd:outline-none',
+          // Subtle scale + translate on open/close (see `popper-in` /
+          // `popper-out`). Radix keeps Content mounted through the exit
+          // animation via `data-state="closed"`. Radix provides
+          // `--radix-select-content-transform-origin` which anchors the
+          // scale to the trigger's corner based on `align` (top-left for
+          // `align="start"`, top-right for `align="end"`).
+          'zd:data-[state=open]:animate-popper-in',
+          'zd:data-[state=closed]:animate-popper-out',
+          className,
+        )}
         {...props}
         style={{
           width: 'var(--radix-select-trigger-width)',
+          transformOrigin: 'var(--radix-select-content-transform-origin)',
           ...style,
         }}
       >
         <Wrapper
           variant="solid"
+          // Override Wrapper's 80% white translucency — a dropdown over
+          // content shouldn't let the content bleed through.
+          style={{ backgroundColor: '#fff' }}
           className="zd:flex zd:flex-col zd:rounded-2xl zd:overflow-hidden"
         >
           <SelectPrimitive.Viewport className="zd:overflow-y-auto zd:max-h-80">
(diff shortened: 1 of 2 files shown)


packages/react-ui/src/components/Select/index.tsx
```diff
@@ -25,6 +25,7 @@ export function SelectTrigger({
       className={cn(
         'zd:inline-flex zd:items-center zd:justify-between zd:gap-2',
         'zd:outline-none zd:cursor-pointer',
+        'zd:data-[state=open]:pointer-events-auto zd:data-[state=open]:cursor-pointer',
```
andrascodes: Two small things on this line:

- `zd:data-[state=open]:cursor-pointer` is redundant — the line above already sets `zd:cursor-pointer` unconditionally, so the open-state variant is a no-op. Can drop it and keep just:
  ```tsx
  'zd:data-[state=open]:pointer-events-auto',
  ```
- The `pointer-events-auto` override is non-obvious — re-asserting it implies you hit a case where Radix set `pointer-events: none` on the open trigger. That's a legit workaround, but a one-line comment naming the behavior it counters would stop someone "cleaning it up" later.

packages/react-ui/src/components/Select/index.tsx
```diff
@@ -76,15 +77,23 @@ export function SelectContent({
         ref={ref}
         position={position}
         sideOffset={sideOffset}
-        className={cn('zd:z-50 zd:max-h-80 zd:outline-none', className)}
+        className={cn(
+          'zd:z-50 zd:max-h-80 zd:outline-none',
+          'zd:data-[state=open]:animate-popper-in',
```
andrascodes, line 85: Entry-only animation — there's no `data-[state=closed]:animate-popper-out`, so the panel snaps shut instantly on close. Totally fine as a design choice (animate in, snap out); just flagging in case a symmetric close was intended. Note that if you do add one, Radix Select doesn't defer unmount for CSS exit animations the way `AnimatePresence` would, so it'd need testing.

packages/react-ui/src/components/Select/index.tsx
```diff
@@ -76,15 +77,23 @@ export function SelectContent({
...
           width: 'var(--radix-select-trigger-width)',
+          transformOrigin: 'var(--radix-select-content-transform-origin)',
           ...style,
         }}
       >
         <Wrapper
           variant="solid"
+          // Override Wrapper's 80% white translucency — a dropdown over
+          // content shouldn't let the content bleed through.
+          style={{ backgroundColor: '#fff' }}
```
andrascodes, line 99: This `#fff` is correct and idiomatic **given today's `Wrapper`**, so nothing to change in this PR — but it's a symptom of a `Wrapper` design smell worth a separate refactor. Capturing it here.

**Problem.** `Wrapper` sets its variant background via an inline `style` (`style={{ backgroundColor, ...rest.style }}`). Inline styles beat any class, so the only way to override the background is *another* inline style — which is why this line (and `Pill`'s disabled `rgba(255,255,255,0.05)`, and `AddressDisplay`'s `rgba(247,245,240,0.4)`) hardcodes a literal with a "inline beats Wrapper's style" comment. Two consequences: it defeats `tailwind-merge` (our `cn` would otherwise let a caller's `bg-*` win last), and it blocks theming (every surface is a hardcoded light-mode literal).

**Solution 1 — background as a utility class (preferred, low-cost, do now in a separate PR).** Map variants to `bg-*` utilities and let `className` win via tailwind-merge:
```tsx
const variantBg: Record<WrapperVariant, string> = {
  ghost: 'zd:bg-white/20',
  soft:  'zd:bg-white/50',
  solid: 'zd:bg-white/80',
}
// cn(base, variantBg[variant], className) — drop the inline backgroundColor, keep rest.style passthrough
```
Then the escapes become plain classes: `SelectContent` → `zd:bg-white`, `Pill` disabled → `zd:bg-white/5`, `AddressDisplay` → `zd:bg-offWhite/40`. Non-breaking — existing inline-style callers keep working during migration.

**Solution 2 — CSS-variable-backed tokens (follow-up, only if dark mode is planned).** Back the palette with CSS vars and add a semantic `surface` token, mirroring how the config already does `var(--zd-spacing)` / `var(--zd-density)`. Then `variantBg` uses `zd:bg-surface/20|50|80` and the surface system re-themes from one place. Solution 1 makes this a drop-in later (swap `white` → `surface` in the one map).

**Recommendation:** Solution 1 as a small standalone PR; Solution 2 only when theming is actually on the roadmap. Not this PR. (Minor aside: `Wrapper`'s unconditional `backdrop-blur-[15px]` is a dead compositing layer behind an opaque fill — worth making conditional while in there.)

andrascodes (approved):
Approving, but I'd really consider doing the refactor of the Wrapper background in a separate PR.

packages/react-ui/src/components/Select/index.tsx
OmarBasem: (animate in, snap out) is the intention

PR #337, opened 2026-07-20
demo(sra): add simulated deposits

OmarBasem:
### Summary

This PR adds simulated deposits to the SRA demo. Refactored from https://github.com/zerodevapp/smart-routing-address-react/pull/5


apps/sra-demo/src/app/globals.css
```diff
@@ -31,21 +37,440 @@ body {
...
+}
+
+@media (max-width: 900px) {
+  .pg__widget {
+    justify-self: center;
+  }
+}
+
+/* ---------- Intro copy ---------- */
+.pg__intro {
```
brtkx: nit: any reason we use css for those not tailwind? I'm not too bothered since it's just a demo, but it would be consistent with styling we use for everything else.

brtkx (approved):

PR #338, opened 2026-07-21
feat(sra): TxnItem component

OmarBasem:
Closes [FS-2440](https://linear.app/offchain-labs/issue/FS-2440/txnitem-component)

### Summary

This PR implements `TxnItem` component

### DIFF
--- a/packages/smart-routing-address-react-ui/src/components/TxnItem/index.tsx
+++ b/packages/smart-routing-address-react-ui/src/components/TxnItem/index.tsx
@@ -1,0 +1,178 @@
+import { cn, Icon, Text } from '@zerodev/react-ui'
+
+export type TxnStatus = 'Routing' | 'Detected' | 'Received' | 'Failed'
+
+export interface TxnItemProps {
+  /** Pre-formatted amount, e.g. `"$248.00 USD"`. */
+  amount: string
+  /** Short address/hash display, e.g. `"0x4d2a…ba99"`. */
+  address: string
+  /** Optional block-explorer URL. Address row becomes a link when supplied. */
+  href?: string
+  /** Pre-formatted relative timestamp, e.g. `"2 mo ago"`. */
+  timestamp: string
+  status: TxnStatus
+  sourceTokenIconUrl?: string
+  sourceChainIconUrl?: string
+  destTokenIconUrl?: string
+  destChainIconUrl?: string
+  className?: string
+}
+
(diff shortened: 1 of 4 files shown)


packages/smart-routing-address-react-ui/src/components/TxnItem/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+  )
+
+  const className =
+    'zd:flex zd:min-w-0 zd:items-center zd:gap-1 zd:text-body3 zd:text-greyScale/50'
+
+  if (href) {
+    return (
+      <a
+        href={href}
+        target="_blank"
```
andrascodes, line 169: Optional a11y: this link opens off-site in a new tab, but its accessible name is just the address (the export icon is `aria-hidden`), so screen-reader users aren't told where it goes or that it opens a new tab. Consider:
```tsx
<a href={href} target="_blank" rel="noreferrer noopener"
   aria-label={`View ${address} on block explorer`}
   className={cn(className, 'zd:hover:text-greyScale')}>
```

packages/smart-routing-address-react-ui/src/components/TxnItem/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+}: {
+  tokenIconUrl?: string
+  chainIconUrl?: string
+  badgeSide: 'left' | 'right'
+  className?: string
+}) {
+  return (
+    <div
+      className={cn(
+        'zd:relative zd:size-11 zd:shrink-0 zd:rounded-xl',
```
andrascodes, line 115: FYI, no action needed: the disc geometry here — 44px well (`zd:size-11`) + centered 34px inner disc (`zd:size-8.5`) — plus the inner-shadow chrome string is now the third copy of a pattern also in `Pill` (`Pill/index.tsx:81`/`87`/`64`) and `PillSkeleton` (`151`/`152`/`145`). The shadow copies have even drifted (the two inset clauses are in reversed order here vs. `Pill`, though visually identical). Not worth extracting today — just a bookmark: if a fourth copy appears, this is a candidate for a shared `<Disc>` primitive (the geometry maps cleanly; the shadow would need a prop since its host element differs).

andrascodes (approved):
LGTM — clean component with genuinely good story + test coverage (link vs. no-link, every status, missing-icon degradation, image rendering all exercised). External-link security is correct and tested, and status is backed by text rather than color alone. Just one optional a11y nit + one FYI below.

PR #339, opened 2026-07-21
feat(sra): pending deposits

OmarBasem:
Closes [FS-2471](https://linear.app/offchain-labs/issue/FS-2471/pending-deposit-ui)

### Summary

This PR implements pending deposit UI

### DIFF
--- a/packages/react-ui/src/components/Screen/index.tsx
+++ b/packages/react-ui/src/components/Screen/index.tsx
@@ -96,7 +96,11 @@ export function Screen({
         <div className="zd:relative zd:z-10 zd:flex zd:flex-1 zd:flex-col zd:min-h-0">
           {topNav}
           <div
-            className="zd:flex zd:flex-1 zd:flex-col zd:min-h-0 zd:overflow-y-auto zd:overflow-x-hidden"
+            // Negative right margin + right padding pushes the scrollbar
+            // flush against the inner card's right edge (as close to the
+            // outer frame as possible without touching the border ring)
+            // while keeping the 16px content gutter intact.
+            className="zd:flex zd:flex-1 zd:flex-col zd:min-h-0 zd:overflow-y-auto zd:overflow-x-hidden zd:-mr-4 zd:pr-4"
             // Scale via --zd-spacing (matches TopNav's scaled height) so the
             // top padding shrinks with the frame — otherwise the fixed 68px
             // eats a disproportionate share at smaller sizes and overflows.
--- a/packages/react-ui/src/components/Screen/index.tsx
+++ b/packages/react-ui/src/components/Screen/index.tsx
@@ -96,10 +96,6 @@ export function Screen({
         <div className="zd:relative zd:z-10 zd:flex zd:flex-1 zd:flex-col zd:min-h-0">
           {topNav}
           <div
-            // Negative right margin + right padding pushes the scrollbar
-            // flush against the inner card's right edge (as close to the
-            // outer frame as possible without touching the border ring)
-            // while keeping the 16px content gutter intact.
             className="zd:flex zd:flex-1 zd:flex-col zd:min-h-0 zd:overflow-y-auto zd:overflow-x-hidden zd:-mr-4 zd:pr-4"
             // Scale via --zd-spacing (matches TopNav's scaled height) so the
             // top padding shrinks with the frame — otherwise the fixed 68px
--- a/packages/smart-routing-address-react-ui/src/components/PendingDeposits/index.tsx
+++ b/packages/smart-routing-address-react-ui/src/components/PendingDeposits/index.tsx
@@ -1,0 +1,129 @@
+import { cn, Text } from '@zerodev/react-ui'
+import type { DepositedToken } from '@zerodev/smart-routing-address'
+import { CHAIN_ICONS, TOKEN_ICONS } from '../../iconAssets'
+import type {
+  DepositStage,
+  EstimatedFee,
+  SmartRoutingAddressConfig,
+} from '../../types'
+import {
+  getDestTokenSymbol,
+  getSourceTokenSymbol,
+  resolveDestChain,
+  sourceTokensFromFees,
+} from '../../utils/config'
+import { getDepositStage } from '../../utils/deposits'
+import { findFeeDataByToken } from '../../utils/fees'
+import {
+  formatDisplayAmount,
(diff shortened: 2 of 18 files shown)


packages/react-ui/src/components/Screen/index.tsx
```diff
@@ -96,7 +96,7 @@ export function Screen({
         <div className="zd:relative zd:z-10 zd:flex zd:flex-1 zd:flex-col zd:min-h-0">
           {topNav}
           <div
-            className="zd:flex zd:flex-1 zd:flex-col zd:min-h-0 zd:overflow-y-auto zd:overflow-x-hidden"
+            className="zd:flex zd:flex-1 zd:flex-col zd:min-h-0 zd:overflow-y-auto zd:overflow-x-hidden zd:-mr-4 zd:pr-4"
```
brtkx, line 99: What's the reason for adding negative margin and padding here?

Btw we've done it before in our team and imo it works quite well - for parts of code that aren't obvious, we could write an inline comment on our own PRs briefly explaining reasons behind a change. Such as this one.

OmarBasem: @brtkx That's for the scroll indicator to be on the edge

Screenshot 2026-07-24 at 9 08 49 AM
Screenshot 2026-07-24 at 9 08 35 AM

brtkx: What if there's no scrollbar. Would the content have uneven horizontal padding then?

packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -132,6 +135,29 @@ export function Deposit({ onQrClick }: DepositProps) {
...
+      setActiveRoute(null)
+      return
+    }
+    setActiveRoute({
+      sourceChainId: source.chain.id,
+      sourceChainName: source.chain.name,
+      token,
+      symbol: sourceSymbol,
+      decimals: feeData.decimal,
+      feeAmount: feeData.fee,
```
andrascodes: `feeAmount` gets written here as `feeData.fee`, which is a **hex** string (`0x3d090`), whereas the demo's `FALLBACK.feeAmount` is a **decimal** string (`250000`) — so the same `ActiveRoute` field carries two different representations depending on the code path.

This matters because `activeRoute` is public API: it's written internally via `setActiveRoute` (from `useSmartRoutingAddressContext`) but read by external hosts through the exported `useSmartRoutingAddress` hook — it's one shared piece of Provider state. Both paths happen to work today only because consumers use `BigInt()`, but the type doc says "atomic units in string" without hinting it's sometimes hex, so any host doing `Number(feeAmount)` / `parseInt(x, 10)` breaks on the hex path. Normalize at the write site:
```ts
feeAmount: BigInt(feeData.fee).toString(), // always decimal atomic units
```

packages/smart-routing-address-react-ui/src/components/PendingDeposits/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <ul className="zd:flex zd:w-full zd:flex-col">
+        {deposits.map((raw) => {
+          const deposit = raw as DepositWithTimestamp
+          const { chainId, token, amount, transactionHash } = deposit.deposit
+          const feeData = findFeeDataByToken(estimatedFees, chainId, token)
+
+          // Source pair: reconstruct the SourceToken so we can look up its
+          // symbol + chain icon the same way the trigger pill does.
+          const source =
+            sourceTokensFromFees(estimatedFees).find(
```
andrascodes: `sourceTokensFromFees(estimatedFees)` is called inside the row `.map`, so it rebuilds the full source-token list once per deposit (O(n·m)). Negligible for a short list, but trivially hoistable above the map:
```ts
const sourceTokens = sourceTokensFromFees(estimatedFees)
// …then inside map: sourceTokens.find(...)
```

packages/smart-routing-address-react-ui/src/utils/format.ts
```diff
@@ -72,3 +72,26 @@ export function formatDuration(seconds: number): string {
   if (seconds < 60) return `~${Math.max(1, Math.round(seconds))} sec`
   return `~${Math.round(seconds / 60)} min`
 }
+
+/** ISO timestamp → compact "N unit ago" ("just now", "2 m ago", "3 h ago",
+ * "5 d ago", "2 mo ago", "1 y ago"). Returns `null` on invalid input. */
+export function formatRelativeTime(
```
andrascodes, line 78: `formatRelativeTime` is untested — it's pure, branchy (7 thresholds + invalid→`null`), and already takes an injectable `now`, so it's ideal for a table test. Worth a small `format.test.ts` block covering the boundaries (`just now`, `s`/`m`/`h`/`d`/`mo`/`y`, invalid input).

andrascodes (approved):
Strong PR — good separation, `PendingDeposits` has solid test coverage (empty / header / row-count / every stage→status), stories exercise all stages, and the semantics (`<section aria-label>`, `<ul>/<li>`) are right. The `getDestTokenSymbol` simplification is a genuine correctness fix — the settlement token is config-fixed, not mirrored from the source — with tests updated to match. No correctness bugs found; nits below.

**FYI (cross-cutting):** the `Screen` change (`zd:-mr-4 zd:pr-4` scrollbar-gutter offset) lands on the shared component, so it affects every `Screen` consumer, not just SRA — worth a glance at other usages to confirm no layout shift.

**Trivial:** the `pages/index.tsx` `renderStep` reformat is pure prettier churn unrelated to the feature; could revert to keep the diff focused.

**Question:** on `Deposit` unmount, `activeRoute` isn't reset, so the last selection lingers in context — intentional (sticky for hosts), or should it clear?

packages/react-ui/src/components/Screen/index.tsx
OmarBasem: No, the scrollbar position is `relative` not `absolute`/`fixed` so the padding stays even

OmarBasem:
> Question: on Deposit unmount, activeRoute isn't reset, so the last selection lingers in context — intentional (sticky for hosts), or should it clear?

Intentional, sticky. `activeRoute` lives on the Provider so hosts (e.g. the demo's MockPanel) can mirror the selection across the widget's own sub-view navigation/pages

PR #341, opened 2026-07-23
feat(sra): fees breakdown

OmarBasem:
Closes [FS-2480](https://linear.app/offchain-labs/issue/FS-2480/fees-breakdown)

### Summary

This PR adds fees breakdown to SRA.

### DIFF
--- a/apps/sra-demo/src/app/components/MockPanel.tsx
+++ b/apps/sra-demo/src/app/components/MockPanel.tsx
@@ -33,6 +33,13 @@ const STEP_LABELS: Record<MockStage, string> = {
   completed: 'Sent — track it in the widget.',
 }
 
+const DOT_COLOR: Record<'idle' | MockStage, string> = {
+  idle: 'bg-muted',
+  pending: 'bg-primary animate-mock-pulse',
+  bridging: 'bg-primary animate-mock-pulse',
+  completed: 'bg-[#6bb04f]',
+}
+
 export function MockPanel({ destChainId }: { destChainId: number }) {
   const { addressState, activeRoute } = useSmartRoutingAddress()
   const route = activeRoute ?? FALLBACK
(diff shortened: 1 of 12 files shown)


apps/sra-demo/src/app/components/MockPanel.tsx
```diff
@@ -108,46 +115,59 @@ export function MockPanel({ destChainId }: { destChainId: number }) {
           : "That doesn't match your deposit address."
 
   return (
-    <div className="pg__wallet">
```
OmarBasem, line 111: Demo CSS classes migrated to tailwind, CSS classes was removed here https://github.com/zerodevapp/zerodev-wallet-sdk/pull/337

packages/smart-routing-address-react-ui/src/hooks/useProviderFees.ts
```diff
@@ -0,0 +1,203 @@
...
+      originChainId: sourceChainId,
+      destinationChainId: destChainId,
+      amount,
+      user: (recipient as Address) || QUOTE_USER,
+      symbol,
+      decimals,
+    }
+
+    const controller = new AbortController()
+    const safe = (p: Promise<FeeBreakdown | null>) => p.catch(() => null)
```
andrascodes: Is this swallowing the errors? Should we be worried about that?

packages/smart-routing-address-react-ui/src/utils/providerFees.ts
```diff
@@ -0,0 +1,298 @@
...
+    // percent = raw/1e18 * 100, kept to 4-decimal percent precision via
+    // BigInt before Number conversion (raw can exceed 2^53).
+    return Number((BigInt(raw) * 1_000_000n) / E18) / 10_000
+  } catch {
+    return null
+  }
+}
+
+/** Across totals are in input-token units; USD only when the token is ~$1 */
+function tokenUsd(
```
andrascodes, line 102: Should we just add a bunch of AI generated tests for these? Like I don't really want to review all the implementations here but if there are tests where the input - output is asserted I'm good with that.
Also maybe some E2E tests or integration tests as well if they're easy to write would be great. 

Or actually I think adding Zod schemas here would be the best choice. Add Zod schemas for all the types, infer the types so you can use that with TS and assert in the fetches that you get the correct data from the APIs. Then this will be super robust and we can be sure that the code, E2E tests...etc breaks if the API is changed in a way that we know that the parsing fails.

packages/smart-routing-address-react-ui/src/hooks/useProviderFees.ts
```diff
@@ -0,0 +1,203 @@
...
+    }),
+  })
+  if (!res.ok) return null
+  const json = await res.json()
+  if (!json || json.error) return null
+  return parseRelayFees(json)
+}
+
+/** Pick the cheaper live route by full quoted cost; else the first available */
+function bestRoute(routes: (FeeBreakdown | null)[]): FeeBreakdown | null {
```
andrascodes, line 118: maybe a unit test for this would be good too? just asking Claude to write it is good enough for me

packages/smart-routing-address-react-ui/src/hooks/useProviderFees.ts
```diff
@@ -0,0 +1,203 @@
...
+      safe(fetchAcross(route, controller.signal)),
+      safe(fetchRelay(route, controller.signal)),
+    ])
+      .then(([across, relay]) => {
+        const chosen = bestRoute([across, relay])
+        if (chosen) {
+          // The USD price of the quoted amount comes from Relay — keep it
+          // even when Across won on fees (used for the min-deposit USD pill
+          // on non-stable tokens).
+          chosen.inputUsd = chosen.inputUsd ?? relay?.inputUsd ?? null
```
andrascodes: Can you get me up to speed on the Relay, Across integrations here? Does SRA in general work with either Relay or Across or both being down? Or can the APIs be down and SRA still work?
Because here we have an absolute dependency of Relay working. Even if Across is the only one working we can't get the USD value without Relay. So I'm just wondering if it's possible to decouple the 2?

andrascodes (changes requested):
Not entirely strict about actually having my comments resolved, so they're more of a discussion starter. But if it's quick to add Zod schema and parsing and tests, I think that goes a long way to just make this code extra robust. But let me know what you think!

packages/smart-routing-address-react-ui/src/hooks/useProviderFees.ts
OmarBasem: Intentional, providers are additive enrichment on top of the SRA execution fee, so a single provider failure should degrade gracefully rather than take down the row. `bestRoute` picks whichever quote we got, if both null, the caller falls back to the SRA fee (documented at lines 117-119).

packages/smart-routing-address-react-ui/src/hooks/useProviderFees.ts
OmarBasem: SRA has no dependency on Relay/Across, the SRA server is authoritative for the address, estimated fee, min/max deposit and polling. Relay + Across are enrichment only: they provide the itemised legs shown when you expand the fee row, plus the provider pill and a better fill-time estimate. If both go down, the fee row collapses to a single "Execution Fee" line equal to the full SRA fee; nothing else changes.

On the Relay coupling for USD,  `inputUsd` is actually redundant, removed it. Across and Relay are now independent.

packages/smart-routing-address-react-ui/src/utils/providerFees.ts
OmarBasem: Good points, added tests and zod schema

andrascodes (approved):
Awesome! Thanks!!!

PR #342, opened 2026-07-23
feat(sra): fees tooltip

OmarBasem:
Closes [FS-2481](https://linear.app/offchain-labs/issue/FS-2481/fees-tooltip)


andrascodes (commented):
Shouldn't we use some library for this too? Like a Radix tooltip?

andrascodes (changes requested):
Just requesting changes for the question on whether a Radix tooltip would make more sense here? The AI comments I'm not too worried about.

OmarBasem:
@andrascodes refactored to using Radix

andrascodes (approved):

PR #344, opened 2026-07-23
feat(react-ui): rewrite ListItem with composition slots

brtkx:
## Summary

Rewrites `ListItem` from config props to slots (`icon` / `subtitle` / `trailing`) + `asChild`. Groundwork for the external-wallets flow, which adds ~a dozen wallet rows the old API couldn't express.

## Why

- Every new row variant needed a new prop (`iconName`/`imageUri`, `badgeProps`, `details`, `chevron`, `alert` — several mutually exclusive). Slots take any node, so variants stop growing the API.
- Wallet rows mix icon sources — runtime strings (announced connectors), the icon set, brand components. One `icon` slot fits all three.
- Download rows are links, not buttons — `asChild` (Radix Slot) renders the row into a consumer `<a>`.

## Changes

- `ListItem`: three ReactNode slots; the row layout (tile, title column, gaps) stays enforced by the component. `subtitle` takes a string (auto-styled) or a node (e.g. `<Badge />`).
- `asChild` via Radix Slot; `children` without `asChild` is a compile error (discriminated union).
- New exported presets: `ListItemIcon` (standard size/color, `className` merges) and `ListItemChevron`.
- Consumers migrated: `SignUp`, `WalletSelection`, `TxGasFees` (details column inlined into `trailing`; its test now stubs only `ListItemIcon` instead of all of `ListItem`).
- Stories/tests rewritten for the slot API, incl. an `AsLink` story and an `asChild` anchor test. Changeset included.

## Breaking

- Removed `ListItem` props: `iconName`, `imageUri`, `badgeProps`, `details`, `chevron`, `alert` (no consumers on main or in any open PR).
- `ListItemProps` is now a type alias — `extends ListItemProps` becomes `& ListItemProps`.

Closes FS-2485

### DIFF
--- a/packages/react-ui/src/components/ListItem/ListItem.stories.tsx
+++ b/packages/react-ui/src/components/ListItem/ListItem.stories.tsx
@@ -80,7 +80,7 @@ export const AsLink: Story = {
     asChild: true,
     children: (
       // biome-ignore lint/a11y/useAnchorContent: the row layout (incl. the title text) is injected into the anchor via Slot
-      <a href="https://metamask.io" target="_blank" rel="noreferrer" />
+      <a href="https://metamask.io" target="_blank" rel="noopener noreferrer" />
     ),
   },
 }
--- a/packages/react-ui/src/components/ListItem/ListItem.stories.tsx
+++ b/packages/react-ui/src/components/ListItem/ListItem.stories.tsx
@@ -1,6 +1,13 @@
 import type { Meta, StoryObj } from '@storybook/react-vite'
 
-import { ListItem, ListItemSkeleton } from './index'
+import { Badge } from '../Badge'
+import { Text } from '../Text'
+import {
+  ListItem,
+  ListItemChevron,
+  ListItemIcon,
+  ListItemSkeleton,
+} from './index'
 
 const meta = {
   title: 'Shared/ListItem',
@@ -10,14 +17,6 @@ const meta = {
   },
   tags: ['autodocs'],
   argTypes: {
-    iconName: {
-      control: 'text',
-      description: 'Icon name to display',
-    },
-    imageUri: {
-      control: 'text',
-      description: 'Image URI to display',
-    },
     title: {
       control: 'text',
       description: 'Main title text',
@@ -25,14 +24,6 @@ const meta = {
     subtitle: {
       control: 'text',
       description: 'Subtitle text',
-    },
-    chevron: {
-      control: 'boolean',
-      description: 'Show chevron icon',
-    },
-    alert: {
-      control: 'boolean',
-      description: 'Alert styling',
     },
   },
   decorators: [
@@ -49,45 +40,48 @@ type Story = StoryObj<typeof meta>
 
 export const WithIcon: Story = {
   args: {
-    iconName: 'wallet',
+    icon: <ListItemIcon name="wallet" />,
     title: 'Wallet',
     subtitle: 'Connected',
-    chevron: true,
+    trailing: <ListItemChevron />,
   },
 }
 
 export const WithDetails: Story = {
   args: {
-    iconName: 'wallet',
+    icon: <ListItemIcon name="wallet" className="zd:text-solarOrange" />,
     title: 'Transaction',
     subtitle: 'Pending',
-    details: {
-      text: '$100.00',
-      subtext: 'USD',
-    },
+    trailing: (
+      <div className="zd:flex zd:flex-col zd:pr-1">
+        <Text className="zd:text-body1">$100.00</Text>
+        <Text className="zd:text-body3 zd:text-greyScale/50 zd:self-end">
+          USD
+        </Text>
+      </div>
+    ),
   },
 }
 
 export const WithBadge: Story = {
   args: {
-    iconName: 'wallet',
+    icon: <ListItemIcon name="wallet" />,
     title: 'Verified Wallet',
-    badgeProps: {
-      text: 'Verified',
-      variant: 'secondary',
-      leadingIcon: 'check',
-    },
-    chevron: true,
+    subtitle: <Badge text="Verified" variant="secondary" leadingIcon="check" />,
+    trailing: <ListItemChevron />,
   },
 }
 
-export const AlertState: Story = {
+export const AsLink: Story = {
   args: {
-    iconName: 'warning',
-    title: 'Action Required',
-    subtitle: 'Review transaction',
-    alert: true,
-    chevron: true,
+    icon: <ListItemIcon name="wallet" />,
+    title: 'Get MetaMask',
+    trailing: <ListItemChevron />,
+    asChild: true,
+    children: (
+      // biome-ignore lint/a11y/useAnchorContent: the row layout (incl. the title text) is injected into the anchor via Slot
+      <a href="https://metamask.io" target="_blank" rel="noreferrer" />
+    ),
   },
 }
 
--- a/packages/react-ui/src/components/ListItem/index.tsx
+++ b/packages/react-ui/src/components/ListItem/index.tsx
@@ -1,7 +1,7 @@
-import type { ButtonHTMLAttributes } from 'react'
+import { Slot } from 'radix-ui'
+import type { ButtonHTMLAttributes, ReactElement, ReactNode } from 'react'
 
 import { cn } from '../../utils/common'
-import { Badge, type BadgeProps } from '../Badge'
 import { Icon, type IconName } from '../Icon'
 import { Text } from '../Text'
 import { Wrapper } from '../Wrapper'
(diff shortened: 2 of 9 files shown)


packages/react-ui/src/components/ListItem/index.tsx
```diff
@@ -35,102 +35,97 @@ export function ListItemSkeleton({ className }: ListItemSkeletonProps) {
...
+        {...(!asChild && { type: 'button' as const })}
         className={cn(
-          'zd:w-full zd:flex zd:flex-row zd:justify-between zd:items-center zd:p-2 zd:transition-colors zd:cursor-pointer',
-          alert
-            ? 'zd:bg-solarOrange/15'
-            : 'zd:hover:bg-white/20 zd:active:bg-white/30',
+          'zd:w-full zd:flex zd:flex-row zd:justify-between zd:items-center zd:p-2 zd:transition-colors zd:cursor-pointer zd:hover:bg-white/20 zd:active:bg-white/30',
           rest.disabled && 'zd:opacity-50 zd:cursor-not-allowed',
         )}
         {...rest}
```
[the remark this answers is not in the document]
brtkx: In that case the dev should pass proper disabled states to both `ListItem` and the child component. There's nothing to be adjusted on our end.

packages/react-ui/src/components/ListItem/index.tsx
```diff
@@ -35,102 +35,97 @@ export function ListItemSkeleton({ className }: ListItemSkeletonProps) {
...
-                className={cn(
-                  'zd:w-6 zd:h-6',
-                  details || alert
-                    ? 'zd:text-solarOrange'
-                    : 'zd:text-greyScale',
-                )}
-              />
-            ) : null}
+          <div className="zd:w-13 zd:h-13 zd:rounded-2xl zd:bg-white zd:flex zd:items-center zd:justify-center zd:shrink-0">
+            {icon}
```
brtkx, line 89: Single icon component replacing mutually exclusive `imageUri` and `iconName`. For exported wallets to work we'd have to add another layer that supports `ReactNode`. This was the main driver for the refactor.

packages/react-ui/src/components/ListItem/index.tsx
```diff
@@ -35,102 +35,97 @@ export function ListItemSkeleton({ className }: ListItemSkeletonProps) {
...
             <Text className="zd:text-body1">{title}</Text>
-            {subtitle && (
+            {typeof subtitle === 'string' ? (
               <Text className="zd:text-body3 zd:text-greyScale/50">
                 {subtitle}
               </Text>
+            ) : (
+              subtitle
             )}
-            {badgeProps && <Badge {...badgeProps} />}
```
brtkx, line 113: Old `subtitle` and `badgeProps` occupy the same slot in the UI. They are now collapsed into a single `subtitle` of type `ReactNode`. The consumer can compose any ReactNode here.

packages/react-ui/src/components/ListItem/index.tsx
```diff
@@ -35,102 +35,97 @@ export function ListItemSkeleton({ className }: ListItemSkeletonProps) {
...
           </div>
         </div>
-        {details ? (
```
brtkx, line 116: `details` and `chevron` occupy the same slot here as well. They are also mutually exclusive, which is not a very clear API to follow.

Replacing it with a single `trailing` of type `ReactNode` - it occupies the same slot and consumer can insert any element here.

packages/react-ui/src/components/ListItem/index.tsx
```diff
@@ -35,102 +35,97 @@ export function ListItemSkeleton({ className }: ListItemSkeletonProps) {
...
 }
+
+export interface ListItemIconProps {
+  name: IconName
+  className?: string
+}
+
+/** Standard leading icon for the tile — pre-sized and colored for the slot;
+ * override via `className` (e.g. `zd:text-solarOrange`). */
+export function ListItemIcon({ name, className }: ListItemIconProps) {
```
brtkx, line 115: `ListItemIcon` and `ListItemChevron` are used quite often, exporting it from here so the consumer doesn't have to write it all the time. Also ensures the UI stays consistent.

packages/wallet-react-ui/src/signing/components/TxGasFees/index.tsx
```diff
@@ -143,11 +144,22 @@ export function TxGasFees({
           <ListItem
             key={item.tier}
             title={capitalizeFirst(item.tier)}
-            iconName={getTierIcon(item.tier)}
-            details={{
```
brtkx, line 147: Removed `details` from the API because it occupies the same slot as the current `trailing`. It is also only used in this part of the codebase, so it can be inlined instead of having its own component in `ListItem`.

packages/react-ui/src/components/ListItem/index.tsx
```diff
@@ -35,102 +35,97 @@ export function ListItemSkeleton({ className }: ListItemSkeletonProps) {
...
+    <Wrapper className={cn('zd:w-full zd:h-17 zd:rounded-2xl', className)}>
+      <Comp
+        {...(!asChild && { type: 'button' as const })}
         className={cn(
-          'zd:w-full zd:flex zd:flex-row zd:justify-between zd:items-center zd:p-2 zd:transition-colors zd:cursor-pointer',
-          alert
-            ? 'zd:bg-solarOrange/15'
-            : 'zd:hover:bg-white/20 zd:active:bg-white/30',
+          'zd:w-full zd:flex zd:flex-row zd:justify-between zd:items-center zd:p-2 zd:transition-colors zd:cursor-pointer zd:hover:bg-white/20 zd:active:bg-white/30',
           rest.disabled && 'zd:opacity-50 zd:cursor-not-allowed',
```
andrascodes, line 82: `disabled` is only wired up for the native `<button>` path. In `asChild` mode it flows through `{...rest}` onto the child and applies the `zd:opacity-50 zd:cursor-not-allowed` styling, but nothing blocks interaction or exposes the state to assistive tech — so a disabled row rendered as a link (e.g. `<ListItem asChild disabled><Link to="…" /></ListItem>`) *looks* disabled while staying fully clickable.

Since `asChild` can render into any element (routing `<Link>`, a temporarily-unavailable nav row, etc.), I'd suggest making `disabled` behave in both paths rather than restricting it:

```tsx
export function ListItem({ /* … */ asChild, children, className, disabled, ...rest }: ListItemProps) {
  const Comp = asChild ? Slot.Root : 'button'
  return (
    <Wrapper className={cn('zd:w-full zd:h-17 zd:rounded-2xl', className)}>
      <Comp
        {...(!asChild
          ? { type: 'button' as const, disabled }
          : { 'aria-disabled': disabled || undefined })}
        className={cn(
          'zd:w-full zd:flex … zd:active:bg-white/30',
          disabled && 'zd:opacity-50 zd:cursor-not-allowed',
          asChild && disabled && 'zd:pointer-events-none',
        )}
        {...rest}
      >
```

This destructures `disabled` out of `rest` (so it's not forwarded as an invalid/no-op attribute onto an `<a>`), keeps the real `disabled` on the native button, and gives the `asChild` path `aria-disabled` + `pointer-events-none` so the state is both announced and actually enforced.

packages/react-ui/src/components/ListItem/index.tsx
andrascodes, line 89: The tile no longer sizes its child, so every caller must remember `zd:w-6 zd:h-6` (baked into `ListItemIcon`, hand-written on `WalletSelection`'s ``). A raw/unsized node in the `icon` slot will render oversized with no guardrail — the JSDoc is the only enforcement. Fine as a trade-off; flagging in case you want the tile to constrain its child (e.g. `[&>*]:zd:w-6 [&>*]:zd:h-6`).

packages/react-ui/src/components/ListItem/ListItem.stories.tsx
```diff
@@ -49,45 +40,48 @@ type Story = StoryObj<typeof meta>
...
-    subtitle: 'Review transaction',
-    alert: true,
-    chevron: true,
+    icon: <ListItemIcon name="wallet" />,
+    title: 'Get MetaMask',
+    trailing: <ListItemChevron />,
+    asChild: true,
+    children: (
+      // biome-ignore lint/a11y/useAnchorContent: the row layout (incl. the title text) is injected into the anchor via Slot
+      <a href="https://metamask.io" target="_blank" rel="noreferrer" />
```
andrascodes: `rel="noreferrer"` here vs `rel="noopener noreferrer"` in the changeset example — just align them. (Functionally equivalent since `noreferrer` implies `noopener`.)

andrascodes (approved):
Clean, well-motivated refactor — the slot API removes the prop explosion and the code is simpler for it. Verified locally: `ListItem` (9/9) and `TxGasFees` (13/13) pass, `react-ui` typecheck is clean, and the removed props have no remaining `ListItem` consumers. One substantive note (`disabled` + `asChild`) and two nits below; none are blocking.

PR #351, opened 2026-07-27
feat(wallet-react-ui): replace AuthConfig with SignUp composition

brtkx:


apps/zerodev-signer-demo/src/app/page.tsx
```diff
@@ -125,7 +136,19 @@ function LandingPageInner() {
...
+                size="md"
+                logo={<ZeroDevLogo variant="mark" tone="color" className="zd:h-8 zd:w-auto" />}
+                renderSignUp={() =>
+                <SignUp>
+                  <SignUp.Passkey />
+                  <SignUp.Divider />
+                  <SignUp.Google />
+                  <SignUp.Email method={getEmailAuthMethod()} />
+                </SignUp>
+              } />
```
OmarBasem: Worth documenting in the demo the different ways we can render the AuthFlow

brtkx: When I think documentation I feel like it should be in wallet ui readme; which I haven't updated and should do. What do you have in mind by documenting in the demo? The demo comment says you could use default `<AuthFlow />` instead and that's all the options.

packages/wallet-react-ui/src/auth/pages/SignUp/context.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const ctx = useContext(SignUpContext)
+  if (!ctx) {
+    throw new Error('SignUp.* components must be rendered inside <SignUp>')
+  }
+  return ctx
+}
+
+/** Report a method's in-flight state into the shared pending registry.
+ * Cleans up on unmount so a removed unit can't leave the page locked. */
+export function useReportPending(id: string, pending: boolean) {
```
andrascodes: Shouldn't we maybe type the `id` with an `AuthMethod` type?
```
type AuthMethod = "email" | "otp" | "google" ...
```

packages/wallet-react-ui/src/auth/pages/SignUp/context.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const { setPending } = useSignUpContext()
+  useEffect(() => {
+    setPending(id, pending)
```
andrascodes: I feel a bit weird about doing this in a `useEffect`. An auth method becomes pending if the user hits a buttton, right? So we should call `setPending` in the `handleSubmit` instead. And we can just reset on unmount in the `SignUpRoot` component.
Further: couldn't we actually just re-use the auth action hooks?
```
const { isPending: isSendOtpPending } = useSendOTP()
const { isPending: isSendMagicLinkPending } = useSendMagicLink()
...
const anyPending = isSendOtpPending || isSendMagicLinkPending || ...
```
Those hooks are Tanstack Query/Mutation hooks so they're designed to be re-used like so. Then you don't need to actually keep `anyPending` in the state at all, you can just derive it from the Tanstack Mutations' state.

andrascodes: Or is it the case that when you add the external wallets on top of this you won't have the same Tanstack Query hooks and you will need to set the `pendingIds` manually?

packages/wallet-react-ui/src/auth/pages/SignUp/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  return (
+    <div className="zd:-mx-4 zd:my-2 zd:flex zd:items-center zd:gap-3">
+      <div className="zd:h-px zd:flex-1 zd:bg-greyScale/30" />
+      <Text className="zd:text-body3">{label}</Text>
+      <div className="zd:h-px zd:flex-1 zd:bg-greyScale/30" />
+    </div>
+  )
+}
+
+function SignUpDefault() {
```
andrascodes: **`SignUp.Default` takes no props — "default page + our terms URL" costs a full re-composition**

Setting a terms/privacy URL was one of the cheapest things `config.auth` supported. On the new API there's no path to it short of hand-composing every unit:

```tsx
// what I want to write
<SignUp.Default termsAndConditionsUrl="https://example.com/terms" />

// what I have to write
renderSignUp={() => (
  <SignUp termsAndConditionsUrl="https://example.com/terms">
    <SignUp.Passkey />
    <SignUp.Divider />
    <SignUp.Google />
    <SignUp.Email />
  </SignUp>
)}
```

That second block is a verbatim copy of `SignUpDefault`'s body that the integrator now owns, so it silently stops matching our canonical page the next time we change it — which the stacked external-wallets PR will, by adding `SignUp.Wallet` rows.

Fix is to forward the root's props:

```tsx
type SignUpRootProps = {
  children: ReactNode
  termsAndConditionsUrl?: string | undefined
  privacyPolicyUrl?: string | undefined
}

function SignUpRoot({ children, termsAndConditionsUrl, privacyPolicyUrl }: SignUpRootProps) { … }

function SignUpDefault(props: Omit<SignUpRootProps, 'children'>) {
  return (
    <SignUpRoot {...props}>
      <SignUpPasskey />
      <SignUpDivider />
      <SignUpGoogle />
      <SignUpEmail />
    </SignUpRoot>
  )
}
```

Call site becomes `renderSignUp={() => <SignUp.Default termsAndConditionsUrl={…} />}` — still composition, no duplicated default.

Worth deciding separately whether bare `<AuthFlow />` should reach this at all: `auth/index.tsx:44` renders `<SignUp.Default />` with no props, so the above only helps consumers who opt into `renderSignUp`. I think that's the right line to hold — terms URLs belong on the components, per point 5 of the doc — just flagging that "default page + terms" still isn't a zero-config case.

packages/wallet-react-ui/src/auth/pages/SignUp/context.tsx
brtkx: I don't disagree with you, but there are some trade-offs if we decide not to use `useReportPending` here. Feel free to counter it again, because I may be missing something in my reasoning. Few things to address:
1. I don't think you can derive pending state for tanstack like that. `useMutation` is per instance, if we call it here it will get a new observer and won't be linked with the other call site even if it's the same query key.
2. We could set those pending states in `handleSubmit` instead, but this way we're not getting the benefits of authentication pending state - we need to build our own. Easier to see it with code example:
```typescript
// what we have now
const { mutateAsync: authenticateOAuth, isPending } = useAuthenticateOAuth()
useReportPending('google', isPending) // <- isPending subscribes to global pending tracking

function handleClick() {
  await authenticateOAuth({ provider: 'google' })
}
```
```typescript
// if we do it in submit handlers
const { mutateAsync: authenticateOAuth } = useAuthenticateOAuth()

function handleClick() {
  setPending('google', false)
  try {
    await authenticateOAuth({ provider: 'google' })
  } finally {
    setPending('google', true)
  }
}
```

I really don't mind going either way. I think both have pros and cons and makes it quite opinionated. But I'm happy to hear more about it, and if you feel strongly about not having it the way its now I'm good to change it.

brtkx: > Or is it the case that when you add the external wallets on top of this you won't have the same Tanstack Query hooks and you will need to set the `pendingIds` manually?

That too tbh, we may need to set pending states ourselves here - although I haven't thought about it when deving it. Good point tho.

packages/wallet-react-ui/src/auth/pages/SignUp/context.tsx
brtkx: Addressed
https://github.com/zerodevapp/zerodev-wallet-sdk/pull/351/changes/036f402aa48c09b63d67e1db81b2e32ca4ab22df

packages/wallet-react-ui/src/auth/pages/SignUp/index.tsx
brtkx: Addressed the root props. Also dropped email type from `SignUp.Email` and moved it to the root as well - so that devs don't have to compose the entire thing if they want to change the method.
https://github.com/zerodevapp/zerodev-wallet-sdk/pull/351/changes/d4e1f06410d5f3ce991aa154a8d87ce349e55c7e

packages/wallet-react-ui/src/auth/pages/SignUp/context.tsx
andrascodes: > I don't think you can derive pending state for tanstack like that. useMutation is per instance, if we call it here it will get a new observer and won't be linked with the other call site even if it's the same query key.

It's funny that there's [this whole thread](https://github.com/TanStack/query/issues/2304) is about people being surprised that `useMutation` doesn't work as `useQuery`. 😂
They recommend `useMutationState` or `useIsMutating`: https://tanstack.com/query/v5/docs/framework/react/reference/useMutationState
So it can be done but it is indeed more cumbersome. Evaluate how you can make the external wallets work with this because that for me is the biggest factor in the decision.

2. Oh that's a good point about the `pending` re-running the effect. I like that.

I think if you could get the `mutationKey` from each login action we have and somehow wrap each external wallet login in similar `useMutation` hooks as well, then you could just have even a central place - like a `usePendingAuthMethods` hook - where you just use `useMutationState` and return the `isLoading` states (or even others) in a centralized place and without duplicating the state. Having said that, I also think now that `useReportPending` is pretty elegant.

packages/wallet-react-ui/src/auth/pages/SignUp/index.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+// tests exercise only the page logic (composition, registry, guard).
+vi.mock('./Passkey', () => ({
+  SignUpPasskey: () => <div data-testid="unit-passkey" />,
+}))
+vi.mock('./Google', () => ({
+  SignUpGoogle: () => <div data-testid="unit-google" />,
+}))
+vi.mock('./Email', () => ({
+  SignUpEmail: () => <div data-testid="unit-email" />,
+}))
```
andrascodes, line 21: **nit: the terms gate has no coverage — not here, and not in E2E**

Non-blocking, but worth a follow-up at least. The risk table says a unit that skips `guardAgreement()` "is a review catch." That's thin for the one control that legally gates auth, and nothing in the suite can back it up:

- This file mocks all three units out (the block below), so the `guardAgreement` test at `:111` proves the *context* gate works — never that a unit calls it. `GuardProbe` is the only caller under test.
- E2E can't cover it either: the demo renders a bare `<SignUp>` with no terms URLs (`apps/zerodev-signer-demo/src/app/page.tsx:145`), so `requiresAgreement` is `false` for every run of `otp.spec.ts` / `magic-link.spec.ts` / `passkey.spec.ts`.

Net effect: deleting `if (!guardAgreement()) return` from any unit today is green across the entire suite. Since the whole point of the compound is that adding units is cheap — and the external-wallets PR adds several — this feels better as a test than as a convention.

Three short tests, one per unit. Sketch for Google:

```tsx
const authenticateOAuth = vi.fn().mockResolvedValue(undefined)

vi.mock('@zerodev/wallet-react', () => ({
  useAuthenticateOAuth: () => ({ mutateAsync: authenticateOAuth, isPending: false }),
  useSendOTP: () => ({ mutateAsync: vi.fn(), isPending: false }),
  useSendMagicLink: () => ({ mutateAsync: vi.fn(), isPending: false }),
  useRegisterPasskey: () => ({ mutate: vi.fn(), isPending: false }),
  useLoginPasskey: () => ({ mutate: vi.fn(), isPending: false }),
}))
vi.mock('../../hooks/useAuth', () => ({ useAuth: () => ({ goToStep: vi.fn() }) }))
vi.mock('../../components/BlobAnimation', () => ({ BlobAnimation: () => null }))

it('does not start OAuth until the terms box is checked', () => {
  render(
    <SignUp termsAndConditionsUrl="https://example.com/terms">
      <SignUp.Google />
    </SignUp>,
  )
  const google = screen.getByRole('button', { name: /google/i })

  fireEvent.click(google)
  expect(authenticateOAuth).not.toHaveBeenCalled()

  fireEvent.click(screen.getByRole('checkbox'))
  fireEvent.click(google)
  expect(authenticateOAuth).toHaveBeenCalledOnce()
})
```

Two notes on making that run: the `@zerodev/wallet-react` mock has to cover all five hooks even for a Google-only test, because `./index` imports all three units to assemble the compound; and these need the real `SignUpFooter` (not the stub above) to get a clickable checkbox.

Worth asserting the highlight side effect (`zd:border-negative`) in at least one of them too, so "blocked by the gate" is distinguishable from "the click did nothing."

I sketched the mock shapes from reading the units rather than running this, so treat the snippet as a starting point, not paste-green.

andrascodes: I'm just posting some of what Claude brings up, so don't take these too literally. But it's a good point I think that the `guardAgreement` could be tested to make sure it triggers if that makes sense to you as well.

apps/zerodev-signer-demo/src/app/page.tsx
brtkx: Updated readme
https://github.com/zerodevapp/zerodev-wallet-sdk/pull/351/changes/c2f66bfc3dbbc21173f3c9a96967477acc2b1639

packages/wallet-react-ui/src/auth/pages/SignUp/Google.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { mutateAsync: authenticateOAuth, isPending } = useAuthenticateOAuth({
+    mutation: {
+      onSuccess: () => {
+        goToStep('authenticated')
+      },
+    },
+  })
+  useReportPending('google', isPending)
+
+  const handleClick = async () => {
```
andrascodes, line 21: **nit: `Google` is the only unit that doesn't guard on `anyPending`**

`Email` (`:29`) and `Passkey` (`:30`) both open with an `anyPending` bail-out; `handleClick` here goes straight to `guardAgreement()`.

Not a live bug — `ListItem` spreads `disabled` onto a native `<button>` (`react-ui/src/components/ListItem/index.tsx:75,84`), so the browser drops the click. But it means this unit's concurrency guard lives in another package's implementation detail rather than in the unit, and it's brittle in two ways worth pre-empting:

- `ListItem`'s `asChild` path swaps the `<button>` for `Slot.Root` and forwards `disabled` to whatever element you hand it. On an `<a>` that attribute is inert and the click lands.
- `anyPending` is derived from the context registry, which lags `isPending` by an effect flush, so the `disabled` attribute is a frame behind. In-handler it's the same lagging value — but belt-and-braces is exactly why `Passkey` carries both.

One line:

```tsx
const handleClick = async () => {
  if (anyPending) return
  if (!guardAgreement()) return
```

Mostly I'd like all three units to read the same way, now that "copy an existing unit" is the documented path for adding one.

packages/wallet-react-ui/src/auth/index.test.tsx
```diff
@@ -61,7 +64,6 @@ vi.mock('./hooks/useAuth', () => ({
     step: mockStep,
     email: null,
     otpId: null,
-    enabledMethods: ['email', 'google', 'passkey'],
     config: null,
```
andrascodes: **nit: the `useAuth` mock has drifted from the hook**

`config` is this PR's leftover — `useAuth` no longer returns it (`hooks/useAuth.ts:15-26`). While you're in here, two more fields are wrong and predate the PR: `onBack` (the hook returns `goBack`) and `setOtpId` (it returns `setOtpSession`). Absent entirely: `otpEncryptionTargetBundle`, `goBack`, `setOtpSession`, `clearOtpSession`.

Nothing breaks today, and the one interaction worth knowing is benign but accidental: since `goBack` is missing it's `undefined`, so `AuthFlow`'s `goBack !== null` guard (`index.tsx:106`) *passes* and spreads `onLeftButtonClick: undefined` — then `TopNav`'s truthy check (`react-ui/src/components/TopNav/index.tsx:45`) hides the button anyway. The result matches the real empty-history case by luck rather than by mock fidelity, and no test in this file asserts on the back button, so changing either guard would stay green.

Cheapest durable fix is to type the mock so drift can't accumulate:

```tsx
useAuth: (): ReturnType<typeof import('./hooks/useAuth').useAuth> => ({ … })
```

That forces every field to be present and correctly named. If it's more ceremony than you want here, just drop `config` and fix the two names.

packages/wallet-react-ui/src/auth/authStoreSlice.test.ts
```diff
@@ -19,43 +9,44 @@ describe('authStoreSlice', () => {
...
-      store.getState().auth.initialize(config)
+      store.getState().auth.initialize()
 
       const { auth } = store.getState()
-      expect(auth.config).toEqual(config)
-      expect(auth.enabledMethods).toEqual(['email', 'google', 'passkey'])
-      expect(auth.step).toBeNull()
+      expect(auth.otpId).toBe('otp-stored')
+      expect(auth.otpEncryptionTargetBundle).toBe('bundle-stored')
+      window.localStorage.removeItem('zerodev:auth:otpSession')
```
andrascodes: **nit: this cleanup belongs in `afterEach`**

If either assertion above fails, `removeItem` never runs and `zerodev:auth:otpSession` leaks into the next test — which is `is a no-op without a stored session` (`:36`), the one test in the file that depends on the key being absent. One real failure becomes two, and the second is a red herring.

No `afterEach` in this file yet, so it needs the import too:

```ts
import { afterEach, describe, expect, it, vi } from 'vitest'

afterEach(() => {
  window.localStorage.clear()
})
```

Then drop line 33.

packages/wallet-react-ui/src/auth/pages/SignUp/context.tsx
andrascodes: ## What the mechanism does

`<SignUp>` enforces mutual exclusion: while one method is authenticating, every other row disables. It's implemented as a `Set` of in-flight ids on the root, and each unit registers itself:

```tsx
// index.tsx:30
const [pendingIds, setPendingIds] = useState<ReadonlySet<AuthMethod>>(new Set())

// index.tsx:56 — the only thing anyone consumes
anyPending: pendingIds.size > 0
```

```tsx
// context.tsx:33
export function useReportPending(id: AuthMethod, pending: boolean) {
  const { setPending } = useSignUpContext()
  useEffect(() => {
    setPending(id, pending)
    return () => setPending(id, false)
  }, [id, pending, setPending])
}
```

Each unit hardcodes its own id: `useReportPending('google', isPending)`, `'email'`, `'passkey'`.

**The key observation:** the only thing derived from that set is `pendingIds.size > 0`. Nobody ever asks *which* method is pending. The ids exist purely as set keys, so that add/remove is idempotent — the identity is write-only information.

## Problem A — the union is closed, so new unit kinds can't register

Commit `036f402` changed `id: string` to `id: AuthMethod`, and `AuthMethod` is a closed three-member union:

```tsx
export type AuthMethod = 'email' | 'google' | 'passkey'   // types.ts:1
```

The external-wallets PR adds `<SignUp.Wallet walletId="metaMask" />` and `<SignUp.MoreWallets />`. Those rows have to participate in mutual exclusion — while a MetaMask connect is in flight, the Google row must disable. So they must call `useReportPending(…)`. But `'metaMask'` isn't assignable to `AuthMethod`, so it won't compile. That PR now has to pick one of:

| Option | Cost |
| --- | --- |
| Widen `AuthMethod` to include wallet ids | It's a public exported type documented in `README.md:156`, and it means "auth method" — wallet ids don't belong in it |
| Add a second union, e.g. `` type PendingId = AuthMethod \| `wallet:${string}` `` | Fine, but it's a change to this file, landing as a merge conflict in the stacked PR |
| `'metaMask' as AuthMethod` | Defeats the point of typing it |

Under the previous `id: string`, PR3 would simply have worked. So the retyping bought type-safety over data nobody reads, and charged for it in the one direction the compound was explicitly designed to extend.

## Problem B — hardcoded ids aren't per-instance, and the union doesn't fix that

This was my original point, and it survives the retyping unchanged. A closed union constrains *which* strings are legal; it does nothing to guarantee two live components don't pick the same one.

Render two of the same unit and they both report under one key. React runs sibling effects in tree order, so the later one wins:

```tsx
<SignUp>
  <SignUp.Google />   {/* A: isPending true  → setPending('google', true)  */}
  <SignUp.Google />   {/* B: isPending false → setPending('google', false) */}
</SignUp>
```

Final set: empty. `anyPending` is `false` while A is genuinely mid-flight, so every row stays enabled and the user can start a second concurrent flow — exactly what the mechanism exists to prevent.

The idempotence guard doesn't help here, because the two writes carry different values:

```tsx
if (prev.has(id) === pending) return prev   // index.tsx:36 — same-id, same-value only
```

There's a matching unmount hazard: B unmounting runs `setPending('google', false)`, clearing the flag A still depends on.

For Google/Email/Passkey, rendering two is contrived — which is why I filed this as forward-looking rather than a bug. But `SignUp.Wallet` is inherently multi-instance: a wallet list is N copies of one component. Whatever scheme PR3 adopts has to be unique per instance, and nothing in the API enforces it — it's a convention its author has to remember.

## Why `useId()` closes both

```tsx
export function useReportPending(pending: boolean) {
  const { setPending } = useSignUpContext()
  const id = useId()                  // unique per instance, stable across renders
  useEffect(() => {
    setPending(id, pending)
    return () => setPending(id, false)
  }, [id, pending, setPending])
}
```

Collisions become impossible by construction; there's no union to widen, so `SignUp.Wallet` just calls `useReportPending(isPending)`. `useId` is React 18+, and the package peers `^18 || ^19`, so it's available.

What you'd give up is the ability to ask "is Google specifically pending?" — but nothing needs that, and if it ever does, each unit already knows its own `isPending` locally without consulting the registry.

## The type is already pushing back on a legitimate use

The test's pending probe had to be renamed from `'probe'` to `'email'` (`index.test.tsx:102`) solely because the union closed. A test probe isn't an auth method, and now it has to pretend to be one — a small signal that the id is being typed as a domain concept when it's really just a key.

## Severity

Nothing is broken on this PR. All three units use distinct hardcoded ids and there's no way to render duplicates through `SignUp.Default`. This is entirely about the cost landing on the stacked PR — which is why it's worth raising now, while this file is still open, rather than as a conflict later.

andrascodes: This is related to me advising you to type the `AuthMethod` above: https://github.com/zerodevapp/zerodev-wallet-sdk/pull/351#discussion_r3664861287

So this is what Claude raised which might be a fair point for the external wallets later.

andrascodes: Btw do we need to know which AuthMethod is actually pending? Can't we just have a `authPending: boolean` state var instead and just set and unset that? We're always just reading `anyPending` anyway.

packages/wallet-react-ui/src/connector.ts
```diff
@@ -74,12 +63,10 @@ export function zeroDevWallet(
...
   const baseFactory = baseZeroDevWallet(params)
-  const store = createStore({ logo: params.config?.logo })
+  const store = createStore()
 
-  // Initialize auth config if provided
-  if (params.config?.auth) {
-    store.getState().auth.initialize(params.config.auth)
-  }
+  // Restore a persisted OTP session so an email flow survives a reload.
+  store.getState().auth.initialize()
```
andrascodes: **Question: was dropping the `config.auth` condition here deliberate, and should this live in `setup()`?**

This now runs on every `zeroDevWallet()` call — and since the connector is constructed at module scope inside `createConfig` with `ssr: true`, that includes the server pass on every request.

It's safe today: `readStoredOtpSession()` opens with `if (typeof window === 'undefined') return null` (`authStoreSlice.ts:12`), so `initialize()` is a no-op on the server. I checked, nothing is broken.

Two reasons I'd still rather it moved:

- The only thing keeping this SSR-safe is a guard in a different module — the call site has none. Anything added to `initialize()` later that touches `sessionStorage` / `matchMedia` / `document` breaks SSR with nothing local to warn you. This file already treats that as a live hazard: `setup()` at `:112` carries an explicit `typeof window === 'undefined'` bail, with a comment about `getProvider()` crashing on the server.
- `setup()` is wagmi's hook for connector side effects, the connector already implements it, and it's client-only by construction. Hydrating the store in the factory body means it runs even for a connector that's never connected, and once per `zeroDevWallet()` call rather than once per connector setup.

Moving the restore into `setup()` — after `await connector.setup?.()`, before the existing SSR bail — would get both for free.

Not blocking; mostly I want to know whether the unconditional call was intentional.

apps/zerodev-signer-demo/src/app/page.tsx
```diff
@@ -9,6 +10,16 @@ import { AppHeader } from './components/AppHeader'
 
 export const dynamic = 'force-dynamic'
 
+// Email auth method choice, toggled by the cogwheel (stored in localStorage).
```
andrascodes: **nit: there's no cogwheel**

Nothing in the demo writes `zd:emailAuthMethod` — I grepped for the key and for any cog/gear/settings component. The only writers are the browser E2E specs (`e2e/browser/otp.spec.ts:48`, `magic-link.spec.ts:43`, `post-auth.spec.ts:75`), so this is purely a test hook.

The inaccuracy came over verbatim from the deleted `wagmi-config.tsx`, so it predates this PR — but the second sentence *was* correctly updated here (the old one claimed changes "trigger a page reload so this re-runs", which stopped being true once the read moved to render time), so it's just this half that got left behind.

Suggest:

```tsx
// Email auth method choice, set by the browser E2E specs (stored in localStorage).
```

andrascodes (approved):
Most of my comments are just ideas for possible improvements (if even that). But this is a big improvement on the current config based system already so approving. Great job!

packages/wallet-react-ui/src/auth/pages/SignUp/context.tsx
brtkx: That's a good point. Let me look into that and get back

packages/wallet-react-ui/src/auth/pages/SignUp/index.test.tsx
brtkx: yeah I wanted to go through tests before merging to make sure there's coverage, good flag. I'll address some

packages/wallet-react-ui/src/auth/pages/SignUp/Google.tsx
brtkx: Yep got the same from my claude audit, I'll address it

packages/wallet-react-ui/src/connector.ts
brtkx: For now that's the only feature so automatic opt-in. When tx signing is added, we will need this to be optional.

christopherferreira9:
All good for QA testing

christopherferreira9 (approved):
LGTM (qa-lab changes)

PR #353, opened 2026-07-27
feat(wallet-react-ui): add more wallets option to SignUp

brtkx:
Renders a button that shows a full list of supported wallets. This opens as a sheet from the bottom of the auth screen.


apps/zerodev-signer-demo/src/app/page.tsx
```diff
@@ -147,6 +147,7 @@ function LandingPageInner() {
                   <SignUp.Divider />
                   <SignUp.Google />
                   <SignUp.Email />
+                  <SignUp.MoreWallets />
```
brtkx, line 150: I will remove this before merge. Just here so you can test. We will probably have a playground demo for us/devs to easily build these live.

packages/wallet-react-ui/src/auth/index.tsx
```diff
@@ -7,14 +7,9 @@ import { ErrorScreen } from './pages/ErrorScreen'
 import { OtpInput } from './pages/OtpInput'
 import { SignUp } from './pages/SignUp'
 import { Verifying } from './pages/Verifying'
-import { WalletSelection } from './pages/WalletSelection'
 import type { AuthStep } from './types'
 import { hasMagicLinkCodeInUrl, stripMagicLinkCodeFromUrl } from './utils/url'
 
-const TITLE_BY_STEP: Partial<Record<AuthStep, string>> = {
```
brtkx, line 14: Removing because "More wallets" is `BottomSheet` now, not its own screen.

andrascodes (approved):

PR #355, opened 2026-07-28
feat(sra): ErrorRetryCard component

OmarBasem:
Closes [FS-2502](https://linear.app/offchain-labs/issue/FS-2502/sra-error-handling)

### Summary

This PR implements the following: 
- `ErrorRetryCard` component
- Wires error handling logic
- Add simulated errors to the demo


packages/smart-routing-address-react-ui/src/hooks/useDepositStatus.ts
```diff
@@ -78,8 +92,13 @@ export function useDepositStatus({
     return () => {
       cancelled = true
       clearInterval(intervalId)
+      pollRef.current = null
     }
   }, [address, enabled, pollingInterval, baseUrl])
 
-  return state
+  const refetch = useCallback(() => {
+    pollRef.current?.()
```
andrascodes, line 106: Can there be a case here where the `poll()` from the `useEffect` is still running and the user clicks a button to call `refetch()` so that fires off another call for `poll()` so now 2 is running concurrently?

andrascodes: Would it make sense to actually replace most the query and polling state with using `useQuery` from TanStack? Or we don't want to add a peer-dep to Tanstack-Query in this package?

andrascodes (approved):
Do take a look at the comments. There might be some things to improve there. But otherwise nothing blocking on my part

packages/smart-routing-address-react-ui/src/hooks/useDepositStatus.ts
OmarBasem: > Can there be a case here where the poll() from the useEffect is still running and the user clicks a button to call refetch() so that fires off another call for poll() so now 2 is running concurrently?

Good point, fixed

> Would it make sense to actually replace most the query and polling state with using useQuery from TanStack? Or we don't want to add a peer-dep to Tanstack-Query in this package?

I think not justifiable at the current state

PR #356, opened 2026-07-28
feat(wallet-react-ui): add SignUp.Wallet unit for pinned wallet buttons

brtkx:
This renders a specific external wallet. For example, if dev wants to render a metamask wallet only, they'd do:
```typescript
<SignUp.Wallet walletId="metamask" />
```


apps/zerodev-signer-demo/src/app/page.tsx
```diff
@@ -145,7 +145,7 @@ function LandingPageInner() {
                 <SignUp emailAuthMethod={getEmailAuthMethod()}>
                   <SignUp.Passkey />
                   <SignUp.Divider />
-                  <SignUp.Google />
+                  <SignUp.Wallet walletId="metamask" />
```
brtkx, line 148: Will revert, just for testing.

packages/wallet-react-ui/src/auth/pages/SignUp/Wallet.tsx
```diff
@@ -0,0 +1,78 @@
+import { Badge, ListItem, ListItemChevron } from '@zerodev/react-ui'
+import { useConnect, useConnectors } from 'wagmi'
+import { useAuth } from '../../hooks/useAuth'
+import { isCancellationError } from '../../utils/isCancellationError'
+import { matchesWallet, WALLET_GUIDE, type WalletId } from '../../walletGuide'
+import { useReportPending, useSignUpContext } from './context'
+
+/** Dedicated row for a single guide wallet: connects it directly when a live
+ * connector claims it, and is a download link for the vendor page otherwise. */
+export function SignUpWallet({ walletId }: { walletId: WalletId }) {
```
OmarBasem, line 10: The wallet list items looks odd, is there Figma designs for this list?

Image

packages/wallet-react-ui/src/auth/pages/SignUp/Wallet.tsx
```diff
@@ -0,0 +1,78 @@
...
+        onError: (err) => {
+          if (!isCancellationError(err)) {
+            setError(err instanceof Error ? err.message : String(err))
+          }
+        },
+      },
+    )
+  }
+
+  return (
```
OmarBasem, line 70: This looks like a component and not a page, it can be placed under `components`

OmarBasem: I see there is actually a list of other components under `SignUp`, I think that's different from the convention we were following. Now we have a list of components under `components` and other components under `pages`. I don't mind co-locating these components, but we need to have a convention on what to place where

brtkx: tbh these are more like a business logic wrapper around `ListItem` and components are more reusable bits. These buttons will never be reused anywhere else except for this SignUp flow.

packages/wallet-react-ui/src/auth/pages/SignUp/Wallet.tsx
brtkx: Resolved, was Omar's local issue

OmarBasem (approved):

andrascodes (approved):

PR #360, opened 2026-07-29
feat(wallet-react-ui): add SignUp.InstalledWallets unit for auto-discovered wallets

brtkx:
This renders a list of all installed wallets. Used if devs want any announced wallet to render as a list. Dev can somewhat control the list with `excludeWalletIds` and `maxWallets`.


OmarBasem (approved):
Check the comment above by Copilot regarding img icon, might be worth adding a default icon

andrascodes (approved):

PR #363, opened 2026-07-30
fix(react): stop retrying expired-session 4xx + drop dead sessions (DPL-662)

SahilVasava:
## Summary

Fixes the expired-session retry noise behind **DPL-662**. Expired session keys were each producing 3–6 identical backend `401`s (bursts of ~5 in 14s), driving the `KmsAppAuthFailureSurge` warnings. Nobody was blocked — it's pure retry noise from users returning to a tab after their session lapsed.

Part of DPL-662.

## Root cause (3 mechanisms; 2 fixed here)

1. **We retry 4xx.** `useAuthenticators` (and wagmi's authed queries) run through TanStack Query with no retry gate → React Query's default `retry: 3` + refetch-on-focus. A `401` from a dead session key is retried 3× and re-fetched on window focus → the 3–6× amplification.
2. **The dead session lingers.** The pre-expiry auto-refresh is a `setTimeout` that only fires while the tab is open. Close the tab across expiry, return → the timer never ran, the expired session is left in storage, and every mount/focus hammers the dead key.
3. **(not in this PR)** A normal `401` from a non-refresh request doesn't proactively clear the session — a backstop for server-side revocation *before* the client-known expiry. Needs transport-level (`onResponse`) plumbing; tracked as follow-up.

## What's here

- **Skip retrying 4xx** — `useAuthenticators` now uses a `shouldRetryRequest` predicate (retries transient/5xx up to twice, never 4xx). Exported `isClientError` + `shouldRetryRequest` so consumers can apply the same gate to their own QueryClient `defaultOptions.queries.retry`.
- **Drop the dead session** — when the session is already expired on load, `clear()` it (instead of leaving it), and re-evaluate on `visibilitychange` so a session that lapsed while backgrounded is refreshed-if-valid / dropped-if-expired on tab return.

Live sessions are unchanged; an expired one now requires re-auth — the same outcome as a failed refresh (which already `clear()`s).

## Notes

- Typecheck + biome clean; existing `provider.test.ts` (14) green; added `query.test.ts` (4) for the retry predicate.
- **Changeset:** needs `pnpm changeset` (patch) before publish — left to run interactively.


OmarBasem (approved):

packages/react/src/provider.ts
```diff
@@ -150,11 +166,14 @@ export function createProvider({
...
 
     destroy() {
-      // Cleanup timer and subscription
+      // Cleanup timer, subscription, and visibility listener
       if (sessionRefreshTimer) {
         clearTimeout(sessionRefreshTimer)
         sessionRefreshTimer = null
       }
+      if (typeof document !== 'undefined') {
+        document.removeEventListener('visibilitychange', onVisibilityChange)
```
andrascodes, line 175: will document cause issues on RN?

andrascodes: oh lol should have read more attentively the line above

andrascodes (approved):

PR #366, opened 2026-07-31
fix(sra): make QR easier to read by scanners

OmarBasem:
Closes [FS-2510](https://linear.app/offchain-labs/issue/FS-2510/address-qrcode-might-be-hard-to-read)


christopherferreira9:
Huge improvement for mobile but decoders still struggle:
- https://www.qr-codes.com/decoder
- https://qrcoderaptor.com/

Given that scanning with the phone was improved, should we proceed and consider this a fix?

OmarBasem:
@christopherferreira9 applied your suggested border fix, should be working smoothly now

christopherferreira9:
@OmarBasem it's working perfectly now

andrascodes (approved):

PR #367, opened 2026-07-31
test(move-e2e-target): make e2e target QA lab

christopherferreira9:
Moves E2E target to QA lab instead of the demo app. All functionality and tests remain the same, no coverage is lost with this transition.

### Main change
This PR drops the need to have 2 apps being built in order to use a specific `projectId` that is configured for either `otp` or `magicLink`. Both secrets are now used when building the app and then `/?authFlavor=<FLAVOR>` will load up the respective <Flavor> where flavor is either `otp` or `magicLink`.

### Improvements
- There's now a new fixture called `authed-session` that allows the reuse of a single session across different spec files. This results in:
  - `export seed phrase` execution time: ~40s reduced to **~3.3s**
  - `export private key ` execution time: ~35s reduced to **~2.2s**
  - `sign typed data` execution time: ~20s reduced to **~1.8s**
  - reduced rate limiting chances as several specs can share the same auth
-  The QA lab now builds one wagmi config per config-key instead of one per mount


brtkx (approved):

SahilVasava (approved):
LGTM overall — nice cleanup. The single-build-with-`?authFlavor=` switch and the `authed-session` reuse are a solid win.

**Coverage is preserved** — test counts are identical vs main (passkey 4/4, post-auth 5/5, otp 2/2, magic-link 1/1, export 2/2); the big line deltas are just the refactor to the new fixture/helpers.

Two non-blocking observations:

1. **`post-auth.spec.ts` uses the shared `authedPage` fixture but performs session-mutating actions (logout/disconnect).** With the worker-scoped shared session, the first logout clears the context, so every following test falls back to a fresh re-login. It self-heals via the `isLabReady` fallback, so it's not incorrect — but it quietly defeats the reuse speedup exactly there and could mask a real regression as "just needed a re-login." Session-lifecycle tests are probably better on an isolated per-test session; `export.spec.ts` (0 mutations) is the ideal fit for the shared fixture.

2. **`providers.tsx` caches config at module scope (`configCache` Map).** Since this is a `'use client'` component that still renders during SSR, that Map (wagmi config + QueryClient) is shared across requests on the server, and it never evicts. Harmless for a single-user QA lab, but worth a comment so the pattern doesn't get copied into a real app.

Nit: the per-chain env RPC override (`NEXT_PUBLIC_SEPOLIA_RPC_URL` etc.) is gone in favor of the project-scoped staging default — intentional and fine, just worth a one-liner in the PR description since it changes the local override channel to URL params only.

christopherferreira9:
Nice catches @SahilVasava ! I've fixed both:
- made the cache browser-only so nothing is shared across SSR requests
- removed the logout step from `post-auth.spec.ts` as we already have coverage for this in the other specs
- added an annotation for when the fixture needs a new auth for visibility in the report

PR #370, opened 2026-07-31
test(e2e-mitm): adds the boilerplate for mocking

christopherferreira9:
Adds the boilerplate for mocking in Playwright tests and (later) manual testing —
  so we can cover negative scenarios where APIs don't return the expected response.

  One set of declarative definitions, two adapters over it:

  - **`routeMocks`** — Playwright request interception (`page.route`). Used by the specs.
  - **`installMockFetch`** — patches `fetch` inside the app, for driving it by hand.

  Matching is shared (`matchMock`, `orderMocks`, `echoJsonRpcId`), so a definition behaves
  the same whichever serves it.

  ### Writing a mock

  ```typescript
  // e2e/mocks/definitions/userWallet.ts
  import type { MockRequest } from '../types.js'

  export const MOCK_WALLET_ADDRESS = '0xdeadbeef00000000000000000000000000001234'

  export const userWallet: MockRequest[] = [
    {
      // Matched against the real destination URL. A RegExp keeps it independent of
      // host and project id, both of which move with `?kms=`.
      url: /\/user-wallet$/,
      method: 'GET',
      response: {
        walletAddresses: [MOCK_WALLET_ADDRESS],
        userId: 'mocked-user-id',
      },
    },
  ]
  ```

  ```typescript
  // e2e/browser/user-wallet.mocked.spec.ts
  import { expect, test } from '@playwright/test'
  import { createNewAccount } from '../helpers/temp-email.js'
  import { loginWithOtp } from '../helpers/ui-login.js'
  import { MOCK_WALLET_ADDRESS, userWallet } from '../mocks/definitions/userWallet.js'
  import { routeMocks } from '../mocks/routeMocks.js'

  test('wallet strip shows the mocked address', async ({ page }) => {
    const mocked = await routeMocks(page, userWallet)

    const account = await createNewAccount()
    await loginWithOtp(page, account.address, account.authToken) // real login

    await expect(page.getByTestId('wallet-address')).toHaveText(
      new RegExp(`^${MOCK_WALLET_ADDRESS}$`, 'i'),
    )
    // A mock matching nothing would leave the real address on screen — a
    // plausible-looking pass. Assert it actually served.
    expect(mocked.hits()).toBeGreaterThan(0)
  })
  ```


e2e/browser/mocked/harness.mocked.spec.ts
christopherferreira9, line 1: This only tests the harness itself without logging in so it remains fast.

e2e/browser/mocked/user-wallet.mocked.spec.ts
christopherferreira9, line 1: Boilerplate test to show how calls are mocked so that responses become deterministic. Will be specially useful for when we include SRA tests.

e2e/browser/otp.spec.ts
```diff
@@ -22,11 +22,6 @@ test.describe('OTP Flow', () => {
     }
   })
 
-  test('should complete OTP login through the UI', async ({ page }) => {
```
christopherferreira9, line 25: this test is redundant as we're also logging in in the next test

e2e/mocks/definitions/userWallet.ts
christopherferreira9, line 1: this is just an example for a mock and is not being used anywhere (yet)

e2e/mocks/routeMocks.ts
brtkx, line 1: Should we write tests for this?

brtkx (approved):

e2e/mocks/routeMocks.ts
christopherferreira9: agree! Just added tests for this

PR #377, opened 2026-08-04
ci: add codecov

christopherferreira9:
This PR integrates codecov into the repo without gating PRs.


SahilVasava (approved):
Verified: action SHA matches the v7.0.0 tag commit, codecov.yml passes the codecov.io/validate check, CODECOV_TOKEN is set on the repo, and the lcov path matches the root vitest coverage output. Informational-only statuses match the PR intent. codecov/patch check already reporting on this PR.

PR #378, opened 2026-08-05
demo(sra): UX enhancements

OmarBasem:
Closes [FS-2544](https://linear.app/offchain-labs/issue/FS-2544/sra-demo-ux-enhancements)

### Summary

This PR adds some UX enhancements to SRA demo:

- When switching to Mainnet ask the your for delivery address directly insted of telling them to set it in settings
- Rename "Advanced Settings" to "Settings" and keep it expanded by default
- Hide "Max slippage" when slippage is not set
- Show window alert when copying SRA address in Simulated mode
- Make 'X' button on SRA component functional, and add "Fund" button to demo
- Keep the delivery address saved when switching between modes
- When refreshing the page while on Mainnet, go to Mainnet (but still ask for delivery address)

Also removes `targetTokenSymbol` prop. Destination token will match the source.

### DIFF
--- a/packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
+++ b/packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
@@ -227,8 +227,10 @@ export function Deposit({
     return out
   }, [srcTokens, selectedTokenType])
 
+  // Only formatted when the consumer explicitly sets a slippage — otherwise
+  // the server picks its own default and there is no meaningful value to show.
   const slippage =
-    typeof config.slippage === 'number' ? formatSlippage(config.slippage) : '—'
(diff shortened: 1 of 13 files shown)


packages/smart-routing-address-react-ui/src/pages/Deposit.tsx
```diff
@@ -464,10 +466,10 @@ export function Deposit({
               <PillRow
                 left={
                   <Pill
-                    label={destSymbol ?? ''}
+                    label={sourceSymbol ?? ''}
```
brtkx: Just a nit, but we have a mix of source and destination logos/symbols for the same component. We should either rename all to destination, or just keep all as source in those places.

brtkx (approved):

PR #380, opened 2026-08-05
SRA: prepare package for release

OmarBasem:
### Summary

This PR prepares the package for release:

- Adds README
- Exclude `**/*.stories.*` and `src/test/**`  in `tsconfig.build.json`
- Some optimizations to reduce bundle size

Fixes the following 2 bugs:

- `useProviderFees` doesn't set loading to false in the catch block. Could cause infinite load
- `PendingDeposits/index.tsx:78-109` - explorer link + token icon vanish whenever the deposited token isn't matched in resolved source tokens

### DIFF
--- a/packages/smart-routing-address-react-ui/package.json
+++ b/packages/smart-routing-address-react-ui/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@zerodev/smart-routing-address-react-ui",
   "version": "0.0.1",
-  "private": true,
   "description": "React hooks, provider, and UI for ZeroDev Smart Routing Address deposits",
   "repository": {
     "type": "git",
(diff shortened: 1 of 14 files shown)


packages/smart-routing-address-react-ui/src/utils/providerFees.ts
```diff
@@ -1,5 +1,8 @@
 import { formatUnits } from 'viem'
-import { z } from 'zod'
+// `zod/mini` ships a tree-shakable, functional API (`z.optional(z.number())`
```
brtkx: nit: I think we can limit inline comments like these - especially the numbers %; these will change as the package grows

packages/smart-routing-address-react-ui/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@zerodev/smart-routing-address-react-ui",
   "version": "0.0.1",
-  "private": true,
```
OmarBasem, line 4: @brtkx are we going to publish npm package or not for now?

brtkx (approved):

PR #383, opened 2026-08-06
fix(core): harden signer and session transitions

SahilVasava:
Stack 1 of 3. Base security changes for Core.

## Summary
- Reject missing, malformed, and zero wallet owners before signing.
- Verify returned signatures against the exact owner and digest across message, transaction, typed-data, userOp, and EIP-7702 routes.
- Preserve EIP-7702 yParity exactly.
- Keep pending device keys isolated until replacement sessions are validated and durably committed.
- Recover or discard interrupted session transitions without exposing mixed key/session state.
- Harden refresh and logout cleanup, including staged bearer-token removal and an explicit force-recovery path.

## Verification
- Full monorepo typecheck passes, including tests.
- 98 unit files and 1,087 tests pass.
- Live staging integration: 9 tests pass.
- Browser E2E coverage is added in stack 3.

## Review focus
Please review signing invariants, pending-key isolation, storage rollback/recovery, JWT-to-device-key binding, and logout failure semantics before marking ready.


packages/core/src/adapters/viem.test.ts
```diff
@@ -45,9 +100,148 @@ describe('toViemAccount owner resolution', () => {
...
+      signMessage,
+    )
+    const account = await toViemAccount({ ...base, client })
+
+    await expect(account.signMessage({ message: 'transfer' })).rejects.toThrow(
+      /recover|owner/i,
+    )
+  })
+
+  it('rejects an attacker signature on every account signing route', async () => {
```
christopherferreira9, line 130: This appears to be the only test covering `signTransaction` in a throw scenario, would it be worth adding an happy path scenario to this in a follow up?
Also, it seems that `serializer` is always defaulting to `signTransaction`, would it be worth adding a scenario that adds a custom `serializer` plus ensuring the two calls happen in `signTransactionInternal`?

packages/core/src/actions/auth/auth.test.ts
```diff
@@ -163,6 +164,37 @@ describe('authenticateWithEmail', () => {
   })
 })
 
+describe('logout', () => {
+  it('stamps the exact Turnkey delete-api-keys activity in the header', async () => {
+    vi.setSystemTime(1_700_000_000_000)
```
christopherferreira9, line 176: nit:  `vi.setSystemTime` here isn't paired with useFakeTimers, and the file has no afterEach — so the clock stays frozen at this time for the remaining tests. It might cause a failure downstream at a later stage.
Would `afterEach(() => vi.useRealTimers())` be worth adding? Mostly so the next time-sensitive test in this file

SahilVasava: Fixed — the logout test now calls `vi.useFakeTimers()` before `setSystemTime`, with `afterEach(() => vi.useRealTimers())` so the frozen clock no longer leaks into later tests.

packages/core/src/adapters/viem.test.ts
SahilVasava: Good idea — tracking the happy-path `signTransaction` test as the follow-up you suggested (positive case: sign a valid tx, assert `recoverTransactionAddress` recovers the owner). The throw-path stays here.

packages/core/src/actions/auth/auth.test.ts
christopherferreira9: awesome!

christopherferreira9 (approved):

packages/core/src/stampers/indexedDbStamper.ts
```diff
@@ -34,12 +59,29 @@ export async function createIndexedDbStamper(): Promise<ApiKeyStamper> {
       pendingKeyPair = keyPair
       return await generateCompressedPublicKeyFromKeyPair(keyPair)
     },
+    async stampPending(payload: string) {
+      if (!pendingKeyPair) throw new Error('No pending key rotation')
```
brtkx: nit
instead of:
```typescript
if (!pendingKeyPair) throw new Error('No pending key rotation')
const signature = await signWithKeyPair(pendingKeyPair, payload)
```
can do:
```typescript
const signature = await signPending(payload)
```

packages/core/src/native/stampers/secureStore.ts
```diff
@@ -1,58 +1,107 @@
...
+      return
+    }
 
+    const publicKey = await SecureStore.getItemAsync(LEGACY_PUBLIC_KEY)
+    const privateKey = await SecureStore.getItemAsync(LEGACY_PRIVATE_KEY)
     if (publicKey && privateKey) {
-      this.publicKeyHex = publicKey
-    } else {
-      await this.resetKeyPair()
+      await this.persistKeyPair({ publicKey, privateKey })
```
brtkx, line 48: What if legacy keys are broken before this patch? The new validation throws. Then dev can't clear stamper (stamper doesn't exist) and can't logout (no sdk). I think then user has to clear storage to fix the app.

Acceptable risk?

packages/core/src/storage/manager.ts
```diff
@@ -20,31 +20,108 @@ export type StorageManager = {
...
+    activePublicKey: string | null,
+  ): Promise<ZeroDevWalletSession | undefined>
+}
+
+type SessionTransition = {
+  sessionData: ZeroDevWalletSession
+  publicKey: string
+}
+
+let mutationTail = Promise.resolve()
```
brtkx, line 44: could we add an inline comment for why the queue is used here

packages/core/src/utils/utils.ts
```diff
@@ -1,43 +1,65 @@
...
  * @returns {PartialBy<ZeroDevWalletSession, "createdAt" | "id" | "stamperType">} - The parsed session.
  */
-export function parseSession(
-  token: string | ZeroDevWalletSession,
-): PartialBy<ZeroDevWalletSession, 'createdAt' | 'id' | 'stamperType'> {
+export function parseSession(token: string | ZeroDevWalletSession): PartialBy<
+  ZeroDevWalletSession,
+  'createdAt' | 'id' | 'stamperType'
+> & {
+  publicKey?: string
```
brtkx: I think `publicKey` is redundant here?

brtkx (approved):

packages/core/src/native/stampers/secureStore.ts
SahilVasava: Good catch — fixed in c873666. `init()` now self-heals instead of throwing: if a stored keypair fails the ownership check it resets to a fresh pair, so a broken/legacy key can no longer brick the app (previously it could neither log out nor reset the stamper without a manual storage wipe). A key that fails validation cannot produce a valid stamp anyway, so this only costs a re-authentication.

Covered both the primary and the legacy-migration paths via an `isUsable` helper, and deliberately kept it scoped to *validation* failures — SecureStore I/O errors still propagate, so a transient read/write failure never discards a good key. Added tests for both self-heal paths plus the transient-read case.

packages/core/src/utils/utils.ts
SahilVasava: You're right — removed in 9e81812. `PartialBy<ZeroDevWalletSession, 'createdAt' | 'id' | 'stamperType'>` already carries `publicKey?` (it's not in the omitted set), so the `& { publicKey?: string }` was a no-op.

packages/core/src/storage/manager.ts
SahilVasava: Added in 9e81812. Documented that `mutationTail` is a process-wide serialization queue: every mutating op runs through `withMutation`, chaining onto the tail so read-modify-write sequences over the shared active-session pointer and session list can't interleave. It's module-level so concurrent managers over the same storage still serialize, and the tail swallows each outcome so one failed mutation never jams the queue.

packages/core/src/stampers/indexedDbStamper.ts
SahilVasava: Done in 9e81812 — extracted a `requirePending()` helper that guards and returns the pending keypair, now shared by `stampPending` and `signPending` (there were two identical guards). Left `commitKeyRotation`'s guard as-is since its message is intentionally more specific ("…to commit").

PR #384, opened 2026-08-06
fix(react): reset wallet state across lifecycle changes

SahilVasava:
Stack 2 of 3. Depends on #383.

## Summary
- Invalidate owner-sensitive account caches when owners or chains change.
- Destroy and clear session providers during disconnect/logout, then recreate them on re-login.
- Prevent stale providers from silently losing automatic refresh after logout and reconnect.
- Reset OAuth preparation state during teardown.
- Cover browser and native OAuth cancellation/cleanup paths.

## Verification
- Full React package and test typecheck passes.
- Provider, connector, store, and OAuth lifecycle tests pass.
- Live browser refresh, reload, signing, and logout coverage is added in stack 3.

## Review focus
Please review provider ownership, cache invalidation boundaries, reconnect behavior, and teardown ordering.

### DIFF
--- a/packages/react/src/connector.test.ts
+++ b/packages/react/src/connector.test.ts
@@ -63,6 +63,15 @@ import { zeroDevWalletCore } from './core/connector.js'
 
 type ConnectorInstance = ReturnType<ReturnType<typeof zeroDevWalletCore>>
 
+function isRefreshProvider(value: unknown): value is { destroy: () => void } {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    'destroy' in value &&
+    typeof value.destroy === 'function'
+  )
+}
+
 function createConnector(mode?: 'EOA' | '4337' | '7702'): ConnectorInstance {
   const factory = zeroDevWalletCore({
     projectId: 'proj-test',
@@ -93,6 +102,17 @@ describe('zeroDevWallet connector — mode branching', () => {
   })
 
   describe('connect()', () => {
+    it('does not create a kernel account when Core has no live session', async () => {
+      const connector = createConnector()
+
+      await expect(connector.connect({ chainId: sepolia.id })).rejects.toThrow(
+        /not authenticated/i,
+      )
+
+      expect(createKernelAccountMock).not.toHaveBeenCalled()
+      expect(createKernelAccountClientMock).not.toHaveBeenCalled()
+    })
+
     it("default mode is '7702' (passes eip7702Account, no ECDSA plugin)", async () => {
       const connector = createConnector()
       await seedEoa(connector)
@@ -146,6 +166,39 @@ describe('zeroDevWallet connector — mode branching', () => {
         '0xcafecafecafecafecafecafecafecafecafecafe',
       ])
     })
+
+    it('retries the whole chain setup after client creation fails', async () => {
+      createKernelAccountClientMock.mockImplementationOnce(() => {
+        throw new Error('bundler client failed')
+      })
+      const connector = createConnector('4337')
+      await seedEoa(connector)
+
+      await expect(connector.connect({ chainId: sepolia.id })).rejects.toThrow(
+        'bundler client failed',
+      )
+      await connector.connect({ chainId: sepolia.id })
+
+      expect(createKernelAccountClientMock).toHaveBeenCalledTimes(2)
+    })
+  })
+
+  describe('switchChain()', () => {
+    it('does not persist an unsupported chain when setup fails', async () => {
+      const connector = createConnector('4337')
+      await seedEoa(connector)
+      // @ts-expect-error - getStore is added in the connector's Properties.
+      const store = await connector.getStore()
+      store.getState().setActiveChainId(sepolia.id)
+      const switchChain = connector.switchChain
+      if (!switchChain) throw new Error('Expected connector.switchChain')
+
+      await expect(switchChain({ chainId: 1 })).rejects.toThrow(
+        'Chain 1 not found in config',
+      )
+
+      await expect(connector.getChainId()).resolves.toBe(sepolia.id)
+    })
   })
 
   describe('getAccounts()', () => {
@@ -166,5 +219,34 @@ describe('zeroDevWallet connector — mode branching', () => {
       const accounts = await connector.getAccounts()
       expect(accounts).toEqual([])
     })
+  })
+
+  it('recreates the refresh provider after disconnect on the same page', async () => {
+    const connector = createConnector()
+    const firstProvider = await connector.getProvider()
+
+    await connector.disconnect()
+
+    const nextProvider = await connector.getProvider()
+    expect(nextProvider).not.toBe(firstProvider)
+  })
+
+  it('stops refresh but preserves retryable state when logout fails', async () => {
+    const connector = createConnector()
+    await seedEoa(connector)
+    // @ts-expect-error - getStore is added in the connector's Properties.
+    const store = await connector.getStore()
+    const wallet = store.getState().wallet
+    if (!wallet) throw new Error('Expected initialized wallet')
+    vi.mocked(wallet.logout).mockRejectedValueOnce(new Error('backend down'))
+    const provider = await connector.getProvider()
+    if (!isRefreshProvider(provider))
+      throw new Error('Expected refresh provider')
+    const destroy = vi.spyOn(provider, 'destroy')
+
+    await expect(connector.disconnect()).rejects.toThrow('backend down')
+
+    expect(destroy).toHaveBeenCalledOnce()
+    expect(store.getState().eoaAccount).toBe(mockEoaAccount)
   })
 })
--- a/packages/react/src/connector.test.ts
+++ b/packages/react/src/connector.test.ts
@@ -1,6 +1,6 @@
 import type { Config } from '@wagmi/core'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { sepolia } from 'wagmi/chains'
+import { mainnet, sepolia } from 'wagmi/chains'
 
 // SDK mocks — hoisted so they're defined before vi.mock() (which is itself
 // hoisted to the top of the file by vitest).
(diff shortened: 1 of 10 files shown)


packages/react/src/connector.test.ts
```diff
@@ -93,6 +102,17 @@ describe('zeroDevWallet connector — mode branching', () => {
   })
 
   describe('connect()', () => {
+    it('does not create a kernel account when Core has no live session', async () => {
```
christopherferreira9, line 105: As the source of truth for sessions now belongs to core, react is still prone to have a stale session via `rehydrate()` so we may be missing a scenario where core has no session but react has a stale session in it. Since `ConnectorCoreParams` takes `persistStorage` we can extend `createConnector` to accept it and seed one:
```typescript
it('erases a stale persisted session when Core has none', async () => {
  const seeded = JSON.stringify({
    state: {
      session: { id: 'stale-from-old-sdk' },
      activeChainId: sepolia.id,
    },
    version: 0,
  })
  const connector = createConnector(undefined, {
    getItem: () => seeded,
    setItem: () => {},
    removeItem: () => {},
  })
  
  // @ts-expect-error - getStore is added in the connector's Properties.
  const store = await connector.getStore()
  
  expect(store.getState().session).toBeNull()
  expect(await connector.isAuthorized()).toBe(false)
})
```

The existing test above is still a useful guard. Worth renaming it to something like `connect() refuses and builds no clients without an owner`?

packages/react/src/connector.test.ts
```diff
@@ -146,6 +166,39 @@ describe('zeroDevWallet connector — mode branching', () => {
...
+      await expect(connector.connect({ chainId: sepolia.id })).rejects.toThrow(
+        'bundler client failed',
+      )
+      await connector.connect({ chainId: sepolia.id })
+
+      expect(createKernelAccountClientMock).toHaveBeenCalledTimes(2)
+    })
+  })
+
+  describe('switchChain()', () => {
```
christopherferreira9, line 230: This looks like a good opportunity to cover the scenario where the sdk does not commit to a new chain when its setup fails:

```typescript
it('does not commit the new chain when its setup fails', async () => {
  // mainnet IS in config, so this gets past the lookup guard and reaches
  // switchActiveChain — which must finish setupChain before it publishes
  // the new activeChainId.
  const connector = createConnector('4337', [sepolia, mainnet])
  await seedEoa(connector)
  // @ts-expect-error - getStore is added in the connector's Properties.
  const store = await connector.getStore()
  store.getState().setActiveChainId(sepolia.id)
  const switchChain = connector.switchChain
  if (!switchChain) throw new Error('Expected connector.switchChain')

  createKernelAccountMock.mockRejectedValueOnce(new Error('setup failed'))

  await expect(switchChain({ chainId: mainnet.id })).rejects.toThrow(
    'setup failed',
  )

  await expect(connector.getChainId()).resolves.toBe(sepolia.id)
})
```

packages/react/src/provider.test.ts
```diff
@@ -294,3 +299,253 @@ describe('wallet_getCapabilities', () => {
...
+      provider.request({
+        method: 'wallet_switchEthereumChain',
+        params: [{ chainId: '0x1' }],
+      }),
+    ).rejects.toThrow('setup failed')
+
+    expect(store.getState().activeChainId).toBe(sepolia.id)
+    provider.destroy()
+  })
+
```
christopherferreira9, line 548: Would it be worth covering `eth_signTypedData_v4` as well since it goes through the same addres validation as `personal_sign` below?

packages/react/src/provider.test.ts
```diff
@@ -294,3 +299,253 @@ describe('wallet_getCapabilities', () => {
...
+      store,
+      config: { projectId: 'proj-test', chains: [sepolia] },
+      chains: [sepolia],
+    })
+    await vi.waitFor(() => expect(wallet.refreshSession).toHaveBeenCalledOnce())
+    await Promise.resolve()
+
+    expect(store.getState().session).toEqual(session)
+    expect(store.getState().eoaAccount?.address).toBe(EOA_ADDRESS)
+    expect(vi.getTimerCount()).toBeGreaterThan(0)
```
christopherferreira9, line 445: `toBeGreaterThan(0)` tells us a timer exists, not that the retry lands inside the 5s cap (https://github.com/zerodevapp/zerodev-wallet-sdk/pull/384/changes#diff-47562da6f5cd265cd7676d55aea78ecf22140c5606c5bc627a743d99d9c32ac7R171). If we widen the 5s to a much higher value this test still passes, would advancing the clock be a stronger check? Something like:
```typescript
await vi.advanceTimersByTimeAsync(5_000)
expect(wallet.refreshSession).toHaveBeenCalledTimes(2)
```

packages/react/src/store.test.ts
christopherferreira9, line 1: Would it worth adding a test where we make sure we never persist session since we removed it from `partialize`? I've added it back locally and the test still passed.

packages/react/src/native/oauth/expoWebBrowser.test.ts
christopherferreira9, line 1: In both tests the browser wins — the deep link fired here is deliberately non-matching, so it never resolves or rejects. That means the Linking branch never delivers a session id in any test. Would a case where it matches and wins the race be worth adding? It'd also cover dismissBrowser(), which is mocked in beforeEach but never asserted — and its comment says it exists for exactly that scenario.
Cancellation looks uncovered too: deleting the added line `if (r.type !== 'success') throw new Error('OAuth cancelled or failed')` leaves the whole react package green locally.

packages/react/src/provider.ts
```diff
@@ -113,27 +116,66 @@ export function createProvider({
...
+        if (store.getState().session?.id !== session.id) return
+        console.log('Session refreshed successfully')
+        store.getState().setSession(newSession || null)
+        store.getState().setIsExpiring(false)
+      } catch (err) {
+        console.error('Session refresh failed:', err)
+        if (store.getState().session?.id !== session.id) return
+        store.getState().setIsExpiring(false)
+
+        if (
```
brtkx, line 151: nit
```typescript
if (err instanceof RestRequestError && err.status === 401)
```

packages/react/src/provider.ts
```diff
@@ -113,27 +116,66 @@ export function createProvider({
...
+          typeof err === 'object' &&
+          err !== null &&
+          'status' in err &&
+          err.status === 401
+        ) {
+          try {
+            await wallet.logout()
```
brtkx, line 158: It may be an acceptable edge case but food for thought:

What if there's an authenticated account but dev breaks / removes their project? If 403 is thrown `wallet.logout()` would throw here and we'd never clear stale auth. Logout is not possible then.

Could also consider giving devs `forceLogout` method, but could be an overkill.

packages/react/src/connector.test.ts
SahilVasava: Added `drops a stale rehydrated session when Core reports none`: seeds a persisted session, inits with core `getSession` -> null, asserts the session is cleared (core is source of truth) while the chain still rehydrates.

packages/react/src/connector.test.ts
SahilVasava: Added `does not commit the new chain when its setup fails`: chain setup throws; asserts `getChainId()` stays on the previous chain and nothing is committed for the target.

packages/react/src/provider.test.ts
SahilVasava: Added a sibling test for `eth_signTypedData_v4` (address as the first param) asserting the same `Invalid from address` validation as `personal_sign`.

packages/react/src/provider.test.ts
SahilVasava: Strengthened: added a `setTimeout` spy and assert `Math.max(...delays) <= 5000`, so it verifies the retry lands within the cap (with 30s left an uncapped retry would schedule 30_000ms and fail), not just that a timer exists.

packages/react/src/store.test.ts
SahilVasava: Added: seeds a real `session` into state and asserts `store.persist.getOptions().partialize(state)` has no `session` key. Fails if `session` were re-added to `partialize`.

packages/react/src/native/oauth/expoWebBrowser.test.ts
SahilVasava: Added a case where `openAuthSessionAsync` never settles and a matching deep link is fired through the expo-linking listener; asserts it resolves with the session id and removes the subscription.

packages/react/src/provider.ts
SahilVasava: `RestRequestError` is not exported from `@zerodev/wallet-core` public entry (internal-only; see utils/query.ts, which duck-types it for the same reason), so `err instanceof RestRequestError` would need a deep import. Kept the structural `status` check with a comment explaining why. Happy to export `RestRequestError` from core if you prefer the instanceof form.

packages/react/src/provider.ts
SahilVasava: The throw is in core `logout`, so the fix landed there (#383, `sahil/sdk-core-security`): on an ambiguous revocation failure (403/5xx/network) it now preserves the credential and returns `false` instead of throwing, so a logout call site does not crash; `logout({ force: true })` still clears locally when the caller knows the remote is gone (LOG-02 unchanged). This refresh path benefits once the stack rebases onto it.

brtkx (approved):

packages/react/src/connector.test.ts
christopherferreira9: great!

packages/react/src/provider.test.ts
christopherferreira9: Looks good!

packages/react/src/store.test.ts
christopherferreira9: Looks good!

packages/react/src/connector.test.ts
christopherferreira9: looks good!

christopherferreira9 (approved):

PR #385, opened 2026-08-06
test(e2e): cover auth delivery and session lifecycle

SahilVasava:
Stack 3 of 3. Depends on #384.

## Summary
- Separate magic-link and plain-OTP projects so delivery-mode tests cannot pass by extracting a code from the wrong email format.
- Exercise fresh OTP login and existing-wallet logout/re-login in the browser.
- Exercise automatic refresh, signing before and after reload, typed-data signing, transactions, logout, passkeys, and wallet export.
- Make integration setup fail fast when either staging project is missing.
- Typecheck test sources in CI and avoid rebuilding demo SDK dependencies twice.

## Required CI secret
ZD_OTP_PROJECT_ID must reference the plain-OTP staging project. Both localhost:3000 and localhost:3001 must be allowed by the respective test projects.

## Verification
- Typecheck and Biome pass.
- 98 unit files and 1,087 tests pass.
- Live staging integration: 6 files and 9 tests pass.
- Browser E2E: all 14 tests pass.
- Publishable-package check passes.

## Review focus
Please verify CI secret/origin configuration and that magic-link and OTP assertions remain strict and project-specific.


e2e/integration/wallet-operations.test.ts
```diff
@@ -69,37 +72,75 @@ describe('Wallet Operations', () => {
...
+      chainId: CHAIN_ID,
+      type: 'eip1559',
+      nonce: 0,
+      gas: 21_000n,
+      maxFeePerGas: 2_000_000_000n,
+      maxPriorityFeePerGas: 1_000_000_000n,
+      to: account.address,
+      value: 0n,
+    })
+    expect(signedTransaction).toMatch(/^0x[0-9a-f]+$/i)
```
christopherferreira9: `/^0x[0-9a-f]+$/i` matches 0x0, so this passes on almost anything.
#383 changes the return here from `serializeTransaction` to `serializer`, so the bytes we hash and the bytes we return now come from two separate calls. If those ever diverge, could we get a signed transaction whose signature doesn't match its own encoding?

`recoverTransactionAddress({ serializedTransaction })` re-derives the hash from the returned bytes and would cover it. Let me know if you think it's worth adding viem as a devDependency on the root of the monorepo.

e2e/playwright.config.ts
```diff
@@ -1,9 +1,21 @@
...
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url))
 const demoAppDir = path.resolve(__dirname, '../apps/zerodev-signer-demo')
+const envPath = path.resolve(__dirname, '../.env')
+if (existsSync(envPath)) loadEnvFile(envPath)
+
+const otpProjectId = process.env.ZD_OTP_PROJECT_ID
+if (!otpProjectId) {
+  throw new Error('ZD_OTP_PROJECT_ID is required for browser OTP tests')
+}
```
christopherferreira9: Tiny little one — this runs on every Playwright invocation, so --list and single-project runs need the var too. I hit it listing just the magic-link project with:
```
$ ZD_OTP_PROJECT_ID= pnpm exec playwright test --list --project=chromium
Error: ZD_OTP_PROJECT_ID is required for browser OTP tests
```

`otpProjectId` is only read at line 66, inside the webServer block that's skipped when CI is set — so in CI it's never actually used, and locally it's only needed to launch the :3001 dev server. Would it work to move the check into that webServer entry, so someone debugging a magic-link failure doesn't need an OTP project id?

e2e/integration/global-setup.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export async function setup() {
+  if (!process.env.ZD_PROJECT_ID || !process.env.ZD_OTP_PROJECT_ID) {
+    throw new Error(
+      'ZD_PROJECT_ID and ZD_OTP_PROJECT_ID are required for staging integration tests',
+    )
+  }
+
+  await waitForBackend(BACKEND_URL)
+  await ping()
```
christopherferreira9, line 13: Like this! 
It now makes the per-file skip machinery unreachable. Every integration spec has the same three guards in beforeAll — try/catch around waitForBackend, try/catch around ping, then the project-id check — each setting a skipReason, with every it() opening on context.skip(!!skipReason, skipReason). Since both helpers return Promise<void> and throw, and you're awaiting them here alongside the env check, can any of those three conditions still reach a test file? 
If not, is it worth deleting them in this PR? Leaving them in reads like the suite can still skip its way to green when it no longer can, which feels like the thing this file exists to prevent. Happy to take it as a follow-up if you'd rather keep this PR smaller.

e2e/helpers/ui-login.ts
```diff
@@ -35,14 +45,45 @@ export async function loginWithOtp(
...
+    authToken,
+    EMAIL_POLL_INTERVAL_MS,
+    EMAIL_POLL_TIMEOUT_MS,
+  )
+  const otpCode = extractOtpCodeFromMagicLinkUrl(emailContent)
+  if (!otpCode) {
+    throw new Error('Magic-link project sent no verification link')
+  }
+
+  await page.goto(`/verify?code=${otpCode}`)
```
christopherferreira9: Small question on the magic-link path. `extractOtpCodeFromMagicLinkUrl` seems to only capture the `code=value` and then the url is rebuilt `/verify?code=` against baseURL. If the project's magic_link_template changed host or path, would this test still go green? Asking because the docstring in `extractOtpCodeFromMagicLinkUrl` mentions `/callback` instead of `/verify`.
If it does point at the demo app, would navigating the emailed URL directly be worth it? That'd make the test cover the link the user actually receives rather than one we construct. If it points somewhere else, it's fine as-is, would be worth adding a comment saying so, since the current code reads like it's clicking the link.

e2e/browser/post-auth.spec.ts
```diff
@@ -64,33 +78,66 @@ test.describe('Post-Auth Operations', () => {
...
+    const initialSessionId = await page.evaluate(() =>
+      localStorage.getItem('@zerodev/active_session'),
+    )
+    if (!initialSessionId) throw new Error('Expected an active browser session')
+
+    await page.evaluate((sessionId) => {
+      const stored = localStorage.getItem(sessionId)
+      if (!stored) throw new Error('Expected the active session record')
+      const session = JSON.parse(stored)
+      session.expiry = Date.now() + 65_000
```
christopherferreira9, line 109: Could we get a comment on why we went for 65s? `SESSION_WARNING_THRESHOLD_MS` is 60s in `provider.ts`, so this lands the refresh ~5s after reload, well inside the 30s `waitForResponse` below.
If that default ever moves, would this fail in a way anyone could diagnose? Raising it past 65s makes it refresh immediately, the test still passes, but it'd be covering the immediate path instead of the scheduled timer, silently. Lowering it to 30s would push the refresh past the `waitForResponse` timeout.  
Seems like this expiry can be set via config, would it be worth setting this on the app side to decouple from the default value?
I'm also good with just adding a comment, the intent behind this test is clear enough.

e2e/integration/wallet-operations.test.ts
SahilVasava: Good catch on both. Switched to `recoverTransactionAddress({ serializedTransaction })` and assert it recovers to `account.address`, so it re-derives the signer from the returned bytes and would catch a serializer/encoding divergence, unlike the hex-shape regex (which also matched 0x0). Added `viem` as a root devDependency as you offered. Passes against real staging signatures in the integration run.

e2e/playwright.config.ts
SahilVasava: Fixed. Dropped the top-level throw; the OTP demo server is now started only when `ZD_OTP_PROJECT_ID` is set, so `--list`, single-project, and magic-link runs no longer require it. `--list --project=chromium` works without it now.

e2e/integration/global-setup.ts
SahilVasava: Done in this PR. Removed the per-file `skipReason`/`context.skip` machinery across all 6 integration specs (global-setup already fails the whole run on backend/email/env). Kept the real setup assignments and dropped the now-unused imports.

e2e/helpers/ui-login.ts
SahilVasava: Now navigates the actual emailed URL via a new `extractMagicLinkUrl`, so a `magic_link_template` host/path change is exercised instead of silently passing. Also corrected the `/callback` to `/verify` docstring. Verified: full browser e2e (incl. magic-link + post-auth) is green against staging.

e2e/browser/post-auth.spec.ts
SahilVasava: Added a comment tying the 65s to `SESSION_WARNING_THRESHOLD_MS` (60s) and the failure modes if that default drifts (covers the immediate-refresh path if raised well past it, times out if lowered below the wait window).

christopherferreira9 (approved):
Looks good for QA

e2e/integration/otp-init-shape.test.ts
```diff
@@ -4,35 +4,21 @@
  * by the production TLS Fetcher key (no `dangerouslyOverrideSignerPublicKey`).
  *
  * Doesn't require email service — only the backend on `upgrade_turnkey` and a
- * test project. Skipped without ZD_PROJECT_ID set.
+ * test project. Skipped without ZD_OTP_PROJECT_ID set.
```
brtkx: nit: not skipped anymore, should say it fails

brtkx (approved):

e2e/integration/otp-init-shape.test.ts
SahilVasava: Fixed — the docstring now says the smoke test fails (not skips) without `ZD_OTP_PROJECT_ID`, since global setup requires it.

PR #395, opened 2026-08-10
refactor(sra): API cleanup

OmarBasem:
This PR cleans up SRA API and slims exports to become the widget plus two companion hooks `useSmartRoutingAddress` and `useCreateSmartRoutingAddress`

### Summary of changes

  Hooks
  - Un-export `useDepositStatus` / `useNewDeposits`
  - Split the companion hook: `useSmartRoutingAddress` is now read-only (`addressState`, `activeRoute`). New `useCreateSmartRoutingAddress` returns `ensureAddress` for pre-creating the address before the
  widget mounts. Updated demo app to showcase `useCreateSmartRoutingAddress` usage.

  Config
  - Removed `estimatedFillTimeSeconds`, we determine and display this ourselves
  - Removed `pollingInterval` rather than reserving it. Re-adding later is non-breaking so deferring costs nothing, and an aggressive-polling knob works against our rate-limited status endpoint; revisit on real demand

  Packaging
  - `@zerodev/smart-routing-address`: changed from peer to regular dependency

Other notes

  - `actions` and `slippage` stay on the provider rather than moving to `<SmartRoutingAddress />`, they're address-creation params, not render props: different values mean a different address for the same recipient.
  - `hasLoaded` kept, not a duplicate of `isLoading`, `isLoading` flips per poll, `hasLoaded` latches after the first response and gates skeleton-vs-error UX


packages/smart-routing-address-react-ui/src/hooks/useCreateSmartRoutingAddress.ts
```diff
@@ -0,0 +1,25 @@
...
+export type UseCreateSmartRoutingAddressResult = {
+  /**
+   * Create the address for the recipient if needed (idempotent per
+   * recipient; a new recipient triggers a fresh creation). Creation params
+   * (`actions`, `slippage`, …) come from the provider config so the address
+   * always matches what the widget displays — there are no per-call
+   * overrides. To create addresses with bespoke params, use
+   * `createSmartRoutingAddress` from `@zerodev/smart-routing-address`.
+   */
+  ensureAddress: (recipient: Address) => Promise<void>
```
brtkx: Should we call it something like `createSmartRoutingAddress`? As a dev I'd expect `ensureAddress` to only validate the address.

packages/smart-routing-address-react-ui/src/hooks/useSmartRoutingAddress.ts
```diff
@@ -1,27 +1,25 @@
...
-  ensureAddress: (recipient: Address) => Promise<void>
-  /**
-   * The route the deposit UI is currently showing (source token + chain +
-   * estimated fee). `null` until the picker has seeded a selection.
+   * The route the widget's deposit UI is currently showing (source token +
+   * chain + estimated fee). `null` until the picker has seeded a selection.
+   * Read-only by design: the widget's picker owns this value; hosts mirror
+   * it (e.g. to drive a companion send panel), they don't set it.
    */
   activeRoute: ActiveRoute | null
```
brtkx, line 12: If dev is not using our UI, do they have a way of setting `activeRoute`? I think right now it's only handled by the UI, so `activeRoute` will always be `null` without the UI?

If that's the case, we should also expose `setActiveRoute`.

apps/sra-demo/src/app/page.tsx
```diff
@@ -440,22 +441,14 @@ export default function Home() {
...
-              </div>
+              <FundReopenCard
+                recipient={recipientReady ? (recipient as Address) : undefined}
+                onOpen={() => setSraOpen(true)}
+              />
             ) : recipientReady ? (
               <SmartRoutingAddress
                 recipient={recipient as Address}
                 onClose={() => setSraOpen(false)}
-                onHelp={() => {
```
brtkx, line 456: I believe we agreed to drop `onHelp` support in general. We could remove it from everywhere: types interface, readme etc.

brtkx (approved):

PR #407, opened 2026-08-13
feat(wallet-react-ui): TxHistory page

OmarBasem:
Closes [FS-2562](https://linear.app/offchain-labs/issue/FS-2562/tx-history-index-page)

### Summary

This PR implements the Tx History index page which includes the list of tx history items.

Currently using mocked data.

Wired in the component in the demo.


packages/wallet-react-ui/src/history/mocks.ts
```diff
@@ -0,0 +1,80 @@
+import type { TxHistoryEntry } from './types'
```
OmarBasem, line 1: temporary mock

packages/wallet-react-ui/src/history/pages/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface TxHistoryProps {
+  /** Called when the top-right × close button is clicked. */
+  onClose: () => void
+  /** Activity feed; defaults to the mock feed until a real source lands. */
+  entries?: TxHistoryEntry[] | undefined
+  /** Fired when a row is tapped. Rows are inert when omitted. */
+  onSelectEntry?: ((entry: TxHistoryEntry) => void) | undefined
+  /** Opens the full history on the ZeroDev Portal; footer hidden when
```
andrascodes: I know the "See full history on Portal" is in the designs. But do you know what that refers to? If no, maybe it's worth it to settle whether we need that before merging this PR.

OmarBasem: It opens the transactions on an external source, but it actually shouldn't be there, removed it

andrascodes (approved):

PR #414, opened 2026-08-18
feat(wallet-ui): WalletConnect support - QA changes

brtkx:
QA found an issue where wallets wouldn't deeplink on mobile automatically. The problem is iOS doesn't allow you to navigate to an app on your phone from browser programmatically - there has to be user action such as click.

This PR moves WalletConnect connection fetching up in the tree, so that it's established before the mobile wallet button is clicked, and users can be navigated to their mobile wallet with the WC connection.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/wallet-react-ui/src/auth/components/WalletSheet/index.tsx
+++ b/packages/wallet-react-ui/src/auth/components/WalletSheet/index.tsx
@@ -7,13 +7,12 @@ import {
   QrCode,
   Text,
 } from '@zerodev/react-ui'
-import { useEffect, useRef, useState } from 'react'
+import { useEffect, useState } from 'react'
 import { useConnect, useConnectors } from 'wagmi'
 import { walletConnectLogo } from '../../brandAssets'
 import { useAuth } from '../../hooks/useAuth'
-import { useWalletConnectPairing } from '../../hooks/useWalletConnectPairing'
+import type { WalletConnectPairing } from '../../hooks/useWalletConnectPairing'
 import { isCancellationError } from '../../utils/isCancellationError'
-import { isMobile } from '../../utils/isMobile'
 import { matchesWallet, type WalletGuideEntry } from '../../walletGuide'
 
 export type WalletSheetProps = {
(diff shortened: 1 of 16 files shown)


packages/wallet-react-ui/src/auth/components/WalletSheet/index.tsx
```diff
@@ -7,44 +7,59 @@ import {
...
 
 export type WalletSheetProps = {
   open: boolean
   onOpenChange: (open: boolean) => void
   /** Absent = generic WalletConnect mode (raw-URI QR, no tabs). */
   wallet?: WalletGuideEntry | undefined
+  /** Page-level pairing (preloaded by the SignUp root) shared by every wallet
+   * surface. Expiry is WalletConnect's own signal: the rejected connect()
+   * surfaces through `error` and the existing Try-again re-pairs. */
+  pairing: WalletConnectPairing
```
brtkx, line 25: Redirecting users programmatically from browser to mobile app on iOS doesn't work - it has to be on click. That's why WC pairing is done outside `WalletSheet` and we only pass the pairing result here.

packages/wallet-react-ui/src/auth/hooks/useWalletConnectPairing.ts
```diff
@@ -14,6 +17,11 @@ export type WalletConnectPairing = {
   uri: string | null
   error: string | null
   retry: () => void
+  /** Wrapped deep link for the one-tap mobile redirect into `wallet`'s app,
+   * or null when the tap should just open the sheet (desktop, claiming
+   * installed connector, no usable URI). An errored pairing — including
+   * WalletConnect's own proposal expiry — never redirects. */
+  deepLinkFor: (wallet: WalletGuideEntry) => string | null
```
brtkx, line 24: Deeplink object, used when clicking specific wallet button in `openWalletSheet`.

OmarBasem (approved):

PR #412, opened 2026-08-17
fix(sra): design review

OmarBasem:
Closes [DES-1](https://linear.app/offchain-labs/issue/DES-1/sra-design-review-adjustments)

### Summary

This PR fixes the following design review points:

- "Minimum deposit" text color
- "Past deposits" button UI
- "Active deposits" UI
- Added a "Copy address" button

Not included in this PR is the changes for the Transaction Details page. Will do it in a separate PR after [TokenSummary](https://github.com/zerodevapp/zerodev-wallet-sdk/pull/408) component is merged

PR #420, opened 2026-08-20
fix(sra): design review and UX polish

OmarBasem:
Closes [FS-2587](https://linear.app/offchain-labs/issue/FS-2587/sra-ui-design-review-and-ux-polish)
### Summary

This PR fixes several design issues in SRA, in addition to a few UX improvements:

- Promote TokenSummary to react-ui
- Update TransactionDetails page using TokenSummary component
- Update some icons to match the designs
- Make estimated fee value clickable, not just the arrow
- Animate arrow icon
- Keep mock controls open when using it
- "Powered by ZeroDev" make it more centered vertically
- Destination chain display as loading while in loading state for consistency


.changeset/design-review-pass.md
```diff
@@ -0,0 +1,44 @@
+---
```
brtkx, line 1: [Claude] This is the #412 changeset with the new bullets appended. #398 already consumed that file, so merging as is re-publishes the whole old entry into both changelogs.

Could you trim it to what this PR changes? By my read that's the TokenSummary promotion, ProgressStep restyle, SelectIcon rotation, TransactionDetails redesign, Received → Delivered, the clickable fee value, and the dest-chain pill skeleton. The "TxnItem detailed variant" and "routing rows spin the rays glyph" bullets are #412 work too.

The DataRow bullet also reads wrong now. After this PR both variants use `info-outline`, not just warning.

packages/smart-routing-address-react-ui/src/components/FeeBreakdown/index.tsx
```diff
@@ -216,3 +216,46 @@ export function FeeBreakdownRows({ breakdown }: { breakdown: FeeBreakdown }) {
...
+  panelId?: string | undefined
+  children: ReactNode
+}) {
+  return (
+    <button
+      type="button"
+      onClick={onToggle}
+      aria-expanded={open}
+      {...(panelId && { 'aria-controls': panelId })}
+      aria-label={open ? 'Hide fee details' : 'Show fee details'}
```
brtkx: [Claude] `aria-label` replaces the button content as its accessible name, so a screen reader hears "Show fee details" and never the fee value. Before this change the value sat outside the button and was still announced as the row value.

A visually hidden label keeps the name intact and the e2e stable with a regex match:

```tsx
{children}
<span className="zd:sr-only">
  {open ? 'Hide fee details' : 'Show fee details'}
</span>
<Icon name="chevronDown" ... />
```

Then `getByRole('button', { name: /Show fee details/ })` at `sra-deposit.spec.ts:282` and `:300`.

packages/react-ui/src/components/Select/index.tsx
```diff
@@ -52,7 +52,13 @@ export function SelectIcon({
   return (
     <SelectPrimitive.Icon
       ref={ref}
-      className={cn('zd:shrink-0', className)}
+      // Radix sets data-state on the trigger; the `in-*` variant reads it
+      // from this descendant, so the chevron turns to face up while the
+      // panel is open and animates back down on close.
+      className={cn(
+        'zd:shrink-0 zd:transition-transform zd:duration-200 zd:in-data-[state=open]:rotate-180',
```
brtkx: [Claude] `in-*` matches any ancestor with `data-state="open"`, and Radix `Dialog.Content` sets that, which is what `BottomSheet` renders. Today's consumers are inline so nothing breaks, but a Select inside a sheet would sit permanently flipped. SRA already has `QrSheet`, so it's one step away.

Scoping to the trigger avoids it:

```tsx
// SelectTrigger (className merges onto Pill via asChild)
className={cn('zd:group/select ...', className)}
// SelectIcon
'zd:group-data-[state=open]/select:rotate-180'
```

A comment naming the constraint is fine too if you'd rather not touch the trigger.

brtkx (commented):
[Claude] Nice cleanup on the fee disclosure and the TokenSummary promotion. Two things I'd fix before merge. The changeset re-ships the #412 entry, and the fee button's `aria-label` hides the value from screen readers. Both inline, plus one optional note on the Select chevron selector.

brtkx (approved):

PR #423, opened 2026-08-24
test(core-integration-boundary): adds initial Core Boundary Integration tests

christopherferreira9:
This PR adds the initial Core boundary tests.

## Core integration layer scope
Core's **outbound** dependencies only. Inbound callers (react, react-ui, the wagmi connector) are boundaries of *those* packages and belong in their own tests. Excluded: chain RPC / bundler / paymaster (transport is injected by the caller, never Core's), and the enclave HPKE envelope

## Integration layer bootstrap
- `vitest.integration.config.ts` — glob `packages/*/src/**/*.integration.test.ts`, happy-dom, no live dependencies
- `vitest.config.ts` — **excludes** the integration glob to avoid having everything running at the same time in the same layer
- `.github/workflows/ci.yml` — an `Integration Tests` job

### Boundaries  covered:
- KMS proxy via `rest.ts`
- Transports via `rest.ts`
- Non-HTTP dependencies
  - Session storage adapter
  - Key store
  - Passkey stamper

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,3 +232,87 @@ jobs:
           name: playwright-report
           path: playwright-report/
           retention-days: 14
+
+  integration-test:
+    name: Integration Tests
+    runs-on: ubuntu-latest
+    # No `if:` guard and no `env:` secrets, deliberately: the integration layer
+    # is in-process and must never need a live service. If a test here ever
+    # needs a secret, it belongs in e2e/backend instead.
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4.3.0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 20
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run integration tests
+        run: pnpm test:integration
+
+  ci-gate:
+    name: All checks passed
+    runs-on: ubuntu-latest
+    needs:
+      - lint-and-format
+      - typecheck
+      - build
+      - test
+      - integration-test
+      - backend-e2e-test
+      - e2e-test
+    if: always()
+    steps:
+      - name: Verify every job succeeded
+        env:
+          # Every job in `needs:` and its result, so this script never has to
+          # name them. Adding or removing a gated job is a one-line edit to
+          # `needs:` above and nothing else.
+          NEEDS: ${{ toJSON(needs) }}
+          # The ONLY per-job policy: jobs allowed to report `skipped`. Both are
+          # guarded off for PRs from forks, so a fork PR could never merge otherwise.
+          # Any other job that reports `skipped` fails the gate — a skip there
+          # means it silently did not run.
+          SKIPPABLE: backend-e2e-test e2e-test
+        run: |
+          fail=0
+
+          # `jq` is preinstalled on GitHub-hosted ubuntu runners.
+          while read -r job result; do
+            case "$result" in
+              success)
+                echo "ok       $job"
+                ;;
+              skipped)
+                case " $SKIPPABLE " in
+                  *" $job "*)
+                    echo "skipped  $job (allowed: guarded off for fork PRs)"
+                    ;;
+                  *)
+                    echo "::error::$job was skipped but is not allowed to skip"
+                    fail=1
+                    ;;
+                esac
+                ;;
+              *)
+                echo "::error::$job did not succeed (got '$result')"
+                fail=1
+                ;;
+            esac
+          done <<EOF
+          $(echo "$NEEDS" | jq -r 'to_entries[] | "\(.key) \(.value.result)"')
+          EOF
+
+          if [ "$fail" -ne 0 ]; then
+            echo "::error::CI gate failed — see the job results above."
+            exit 1
+          fi
+          echo "All gated jobs passed."
--- a/packages/core/src/actions/wallet/userOperationPayload.integration.test.ts
+++ b/packages/core/src/actions/wallet/userOperationPayload.integration.test.ts
@@ -1,0 +1,216 @@
+/**
+ * Boundary between Wallet Core and KMS.
+ *
+ * Nothing in Core or the viem adapter calls `sign/user-operation`. Only an app
+ * using the client gets there, so the test calls the action directly.
+ * `computeDataPayloadHash` is specific to this endpoint. `sendSigningRequest`
+ * and `buildTurnkeyPayload`'s address guard are shared with the other signing
+ * paths.
+ */
+import { type Hex, keccak256, recoverAddress, toHex } from 'viem'
+import { privateKeyToAccount } from 'viem/accounts'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { rest } from '../../client/transports/rest.js'
+import type { Client } from '../../client/types.js'
+import type { Stamper } from '../../stampers/types.js'
+import { signUserOperation } from './signUserOperation.js'
+
+const BASE = 'https://kms.example.invalid/api/v1'
+const SIGN_URL = `${BASE}/proj/sign/user-operation`
+
+/** Publicly known test keys. OWNER is the wallet; IMPOSTOR is anyone else. */
+const OWNER = privateKeyToAccount(`0x${'11'.repeat(32)}` as Hex)
+const IMPOSTOR = privateKeyToAccount(`0x${'33'.repeat(32)}` as Hex)
+
+/** A plausible packed user operation. Content is opaque to Core. */
+const USER_OP = '0xdeadbeef'
+
+const stamper: Stamper = {
+  stamp: async () => ({
+    stampHeaderName: 'X-Stamp',
+    stampHeaderValue: 'stamp',
+  }),
+  clear: async () => {},
+}
+
+type SigningRequest = {
+  unsignedUserOperation: string
+  chainId: number
+  encoding: string
+  turnkeyPayload: { parameters: { signWith: string; payload: string } }
+}
+
+function driven(
+  answer?: (correct: Hex) => unknown,
+  signer = OWNER,
+): { client: Client; sent: SigningRequest[]; urls: string[] } {
+  const sent: SigningRequest[] = []
+  const urls: string[] = []
+  vi.stubGlobal(
+    'fetch',
+    vi.fn(async (url: string | URL | Request, init?: RequestInit) => {
+      urls.push(String(url))
+      const body = JSON.parse(String(init?.body)) as SigningRequest
+      sent.push(body)
+      const correct = await signer.sign({
+        hash: `0x${body.turnkeyPayload.parameters.payload}` as Hex,
+      })
+      return new Response(
+        JSON.stringify(answer ? answer(correct) : { signature: correct }),
+        { status: 200, headers: { 'content-type': 'application/json' } },
+      )
+    }),
+  )
+  const client = {
+    request: rest(BASE, {
+      apiKeyStamper: stamper,
+      passkeyStamper: stamper,
+    }).request,
+    apiKeyStamper: stamper,
+  } as unknown as Client
+
+  return { client, sent, urls }
+}
+
+const PARAMS = {
+  organizationId: 'suborg-1',
+  projectId: 'proj',
+  token: 'tok',
+  address: OWNER.address,
+  chainId: 1,
+  encoding: 'hex' as const,
+  unsignedUserOperation: USER_OP,
+}
+
+const outcomeOf = (p: Promise<unknown>) =>
+  p.then(
+    (value) => ({ ok: true, value }) as const,
+    (error: unknown) => ({ ok: false, error }) as const,
+  )
+
+/** 65 bytes exactly. Asserted alongside recovery, never instead of it. */
+const SIGNATURE = /^0x[0-9a-f]{130}$/i
+
+/** Attributable = an Error that is not a raw deref and names the signature. */
+const blamesTheSignature = (error: unknown) =>
+  error instanceof Error &&
+  !(error instanceof TypeError) &&
+  /signature|signing|owner/i.test(error.message)
+
+afterEach(() => {
+  vi.unstubAllGlobals()
+})
+
+describe('user operation: a signature the caller can broadcast', () => {
+  const encodings: ['hex' | 'utf8', string, Hex][] = [
+    ['hex', 'the bytes the hex decodes to', keccak256(USER_OP)],
+    ['utf8', 'the bytes of the string itself', keccak256(toHex(USER_OP))],
+  ]
+
+  for (const [encoding, what, expected] of encodings) {
+    it(`signs ${what} when the encoding is ${encoding}`, async () => {
+      const { client } = driven()
+
+      const signature = await signUserOperation(client, { ...PARAMS, encoding })
+
+      expect(signature).toMatch(SIGNATURE)
+      await expect(recoverAddress({ hash: expected, signature })).resolves.toBe(
+        OWNER.address,
+      )
+    })
+  }
+
+  it('hashes the user operation the caller passed, not something else', async () => {
+    const { client, sent } = driven()
+
+    await signUserOperation(client, { ...PARAMS, encoding: 'hex' })
+
+    expect(`0x${sent[0].turnkeyPayload.parameters.payload}`).toBe(
+      keccak256(USER_OP),
+    )
+  })
+
+  it('hashes the raw string when the encoding is utf8', async () => {
+    const { client, sent } = driven()
+
+    await signUserOperation(client, { ...PARAMS, encoding: 'utf8' })
+
+    expect(`0x${sent[0].turnkeyPayload.parameters.payload}`).toBe(
+      keccak256(toHex(USER_OP)),
+    )
+  })
+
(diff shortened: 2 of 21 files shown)


packages/core/src/storage/sessionTransition.integration.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    expect(mem.store.get(ALL_SESSIONS_KEY)).toBe(before.get(ALL_SESSIONS_KEY))
+    expect(mem.store.get(ACTIVE_SESSION_KEY)).toBe(
+      before.get(ACTIVE_SESSION_KEY),
+    )
+    await expect(manager.getActiveSession()).resolves.toMatchObject({
+      id: 'session_existing',
+    })
+  })
+
+  it('does not report a usable session when the storage read fails', async () => {
```
christopherferreira9, line 226: Question here:
This test asserts only that a throwing storage read must not produce a usable session. It doesn't assert that the read rejects, and it doesn't assert that it resolves to "no session". Should rejected reads result in an empty session? A corrupt journal tolerates bad data, `parseTransition` returns `null` and the entry is discarded but a read that throws propagates all the way out, and `createZeroDevWalletCore` then rejects, so the wallet is unconstructable

packages/core/src/core/registerMidFlowFailure.integration.test.ts
```diff
@@ -0,0 +1,233 @@
...
+    await core.auth(register)
+
+    expect(kms.creates()).toHaveLength(1)
+    expect(kms.logins()).toHaveLength(1)
+    await expect(core.getSession()).resolves.toBeDefined()
+  })
+})
+
+describe('passkey registration: KMS fails between the two calls', () => {
+  it('surfaces the failure and leaves no session, having already created the account', async () => {
```
christopherferreira9, line 203: Question: What is the intended behavior when the second call fails? Do we want to support having a `retry` mechanism or reconcile here or is start-over the intended behavior?
This `describe` blocks explores the possibility of KMS not being available at the time of login, right after registration happens. From the code it looks like the path is for the user to start over by either registering again or logging in reusing the same Passkey (taking passkey as an example).

packages/core/src/core/loginWalletIdentity.integration.test.ts
```diff
@@ -0,0 +1,389 @@
...
+    await account.signMessage({ message: 'hello' })
+
+    const sign = kms.lastOf('sign/message') as {
+      body: { turnkeyPayload: { organizationId: string } }
+    }
+    expect(sign.body.turnkeyPayload.organizationId).toBe(WALLETS[1].subOrgId)
+    expect(sign.body.turnkeyPayload.organizationId).not.toBe('parent-org')
+  })
+
+  it('lands on a different wallet across a logout and a fresh login, with no error either time', async () => {
```
christopherferreira9, line 302: Question:  should login refuse when a session is already active, or is silently switching accounts intended?

The registration flow seems to guard this by throwing `Passkey registration requires an unauthenticated wallet; logout before registering a new wallet.` but login does not follow the same approach. A second login over a live session ends up cleaning the first session through `commitSessionTransition` . 
This test goes through a `logout()` rather than a bare second login on purpose.

packages/core/src/client/kmsFailureFidelity.integration.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+afterEach(() => {
+  vi.unstubAllGlobals()
+})
+
+describe('KMS boundary: HTTP failures stay distinguishable', () => {
+  const cases = [
+    { status: 401, label: 'rejected credential — re-authenticate' },
+    { status: 403, label: 'forbidden — do not retry' },
+    { status: 429, label: 'throttled — back off and retry' },
```
christopherferreira9, line 58: Question: should `rest.ts` retry a 429 or 5xx, or is one attempt then throw intended?
The label here states _back off and retry_ but the SDK does not follow it so the test is naming something an expectation the code doesn't implement which in my opinion is ok.
It seems to matter the most on passkey registration, which makes two calls. A 429 on the second call leaves the account created remotely with no local session.

packages/core/src/adapters/walletAddressResolution.integration.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    for (const account of [accountA, accountB]) {
+      expect(account.type).toBe('local')
+      expect(typeof account.signMessage).toBe('function')
+      expect(typeof account.signTransaction).toBe('function')
+      expect(typeof account.signTypedData).toBe('function')
+    }
+  })
+
+  it.fails(
+    'refuses to build an account when KMS returns more than one address',
```
christopherferreira9, line 184: Question: can one sub-org return more than one wallet address?
This test assumes Core should refuse KMS returning two accounts since it has no way to know which one is meant. `toViemAccount` takes `walletAddresses[0]` and signs with it so with response from KMS with two valid `walletAddresses`, one is silently picked.
If it can't happen, this is a guard against a future change, if it can, this is already live so a user's address would depend on response ordering. 
These integration tests do not assume behavior from boundaries.

packages/core/src/adapters/authorizationPayload.integration.test.ts
```diff
@@ -0,0 +1,452 @@
...
+    })
+
+    await expect(recoverAuthorizationAddress({ authorization })).resolves.toBe(
+      OWNER.address,
+    )
+    expectWellFormedSignature(authorization)
+  })
+})
+
+describe('7702: nothing to authorize', () => {
```
christopherferreira9, line 360: Question: should `signAuthorization` validate the delegation target before even reaching KMS?
This block covers the case where there's no target which is a guarded by `!authAddress` but a malformed target isn't covered because it isn't refused. `signAuthorization({ address: '0xdead', chainId: 1, nonce: 3 })` returns a signature and costs a KMS round trip to then have the chain reject the authorization.

packages/core/src/core/logoutRevocation.integration.test.ts
```diff
@@ -0,0 +1,328 @@
...
+    ['the key list is null', () => ({ sessionKeys: null })],
+    ['the key list is absent', () => ({})],
+    [
+      'the entry has no Turnkey id',
+      (activeKey) => ({ sessionKeys: [{ ApiKey: activeKey }] }),
+    ],
+  ]
+
+  for (const [label, authenticators] of cases) {
+    it(`refuses and keeps local credentials when ${label}`, async () => {
```
christopherferreira9, line 232: Question: should `logout()` tell the caller why it refused?
Currently, the message _Current session key was not found in the authenticated key list; refusing to erase local credentials without confirmed remote revocation._ isn't reachable by any caller as it is thrown inside the try catch so since this isn't a rest request error `credentialIsRejected` is false (https://github.com/zerodevapp/zerodev-wallet-sdk/blob/main/packages/core/src/core/createZeroDevWalletCore.ts#L614-L616) . This results in the app not being able to distinguish a 403 from a network failure.

packages/core/src/core/oauthSessionAbsent.integration.test.ts
```diff
@@ -0,0 +1,309 @@
...
+
+      await expect(core.auth(oauth)).resolves.toBeDefined()
+
+      await expect(core.getSession()).resolves.toBeUndefined()
+      expect(api.activeKey()).toBeNull()
+      expect(api.pendingKey()).toBeNull()
+    })
+  }
+
+  it('returns the payload unchanged, since it is the only signal', async () => {
```
christopherferreira9, line 220: Question: should `auth({ type: 'oauth' })` resolve successfully when the response carries no session?
The oauth branch return early when data.session is absent so `auth()` resolves, the user is not logged in and the returned payload is the only indication. None of these tests asserts that it should reject since that is the question itself.

packages/core/src/core/sessionReconciliation.integration.test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    const core = await buildCore(store, `0x${KEY_LIVE.toUpperCase()}`)
+
+    await expect(core.getSession()).resolves.toMatchObject({
+      id: 'session_live',
+    })
+  })
+})
+
+describe('core reconciliation on construction: stores disagree', () => {
```
christopherferreira9, line 162: Question: is clearing every session right for any key disagreement or only for a crashed transition?
When the active key is compared with the restores session's publicKey, `clearAllSessions()` is called if they differ and the surrounding catch does the same. This same branch fires when a key store was simply evicted (browser clearing IndexedDB after 7 days or a device restore bringing back session storage without the key material. Those sessions are not corrupt, just orphaned and this results in all being cleared and the user is logged out with no signal.

packages/core/src/actions/wallet/userOperationPayload.integration.test.ts
```diff
@@ -0,0 +1,216 @@
...
+    const { client, sent } = driven()
+
+    await signUserOperation(client, { ...PARAMS, encoding: 'utf8' })
+
+    expect(`0x${sent[0].turnkeyPayload.parameters.payload}`).toBe(
+      keccak256(toHex(USER_OP)),
+    )
+  })
+
+  it('sends the operation, chain and encoding on to the backend', async () => {
```
christopherferreira9, line 143: Question: is a user operation signature meant to be chain specific?
This test asserts `chainId` reaches the backend in the body but does not assert anything about the hash. Tried with different `chainIds` as payloads and the signed payload is identical every time meaning that one signature is valid on every chain.

packages/core/src/actions/wallet/userOperationPayload.integration.test.ts
```diff
@@ -0,0 +1,216 @@
...
+      const signature = await signUserOperation(client, { ...PARAMS, encoding })
+
+      expect(signature).toMatch(SIGNATURE)
+      await expect(recoverAddress({ hash: expected, signature })).resolves.toBe(
+        OWNER.address,
+      )
+    })
+  }
+
+  it('hashes the user operation the caller passed, not something else', async () => {
```
christopherferreira9, line 123: Question: should an empty or malformed user operation be refused before it is hashed?
This circles back to other questions in this PR, does Core validate a caller's input before calling KMS, or is that the caller's job? Nothing is asserted either way here.

packages/core/src/adapters/authorizationPayload.integration.test.ts
```diff
@@ -0,0 +1,452 @@
...
+    it(`asks KMS to sign the pre-image it sends for ${label}`, async () => {
+      const { signAuthorization, sent } = await loggedIn(correctly)
+
+      await signAuthorization({ address: DELEGATE, chainId, nonce })
+
+      // Nothing downstream compares the two, so this is the only guard against
+      // the hand-built pre-image drifting from viem's `hashAuthorization`. The
+      // `both zero` case cannot see a chainId/nonce transposition (identical at
+      // 0); the other three can, hence the whole table.
+      expect(preimageOf(sent[0])).toBe(signedHashOf(sent[0]))
```
christopherferreira9, line 285: Question: should `signTypedData` check that `typedDataHash` is the hash of `unsignedTypedDataV4`?

packages/core/src/utils/walletExport.integration.test.ts
```diff
@@ -0,0 +1,347 @@
...
+
+  it.fails('does not wait forever on an unanswered export', async () => {
+    // A fetch that never settles, with fake timers so the clock is advanced
+    // rather than waited on. Neither export helper passes an abort signal, where
+    // `rest.ts` aborts a KMS call after 10 s.
+    vi.useFakeTimers()
+    vi.stubGlobal(
+      'fetch',
+      vi.fn(() => new Promise<Response>(() => {})),
+    )
```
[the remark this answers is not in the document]
christopherferreira9: good catch, this is now fixed.

packages/core/src/adapters/walletAddressResolution.integration.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      // Today: `TypeError: Cannot read properties of undefined (reading '0')`.
+      const error = await shapeViolation({}).catch((e: unknown) => e)
+
+      expect(error).toBeInstanceOf(Error)
+      expect(error).not.toBeInstanceOf(TypeError)
+      expect((error as Error).message).toMatch(/wallet/i)
+    },
+  )
+
+  it.fails(
```
brtkx, line 135: I feel like we should only use `fails` for actual expected failures. For tests to be implemented we could use [todo](https://vitest.dev/api/test.html#test-todo)?

.github/workflows/ci.yml
brtkx, line 1: Readme says the integration layer has no tests yet. This changes with this PR - should we adjust it?

packages/core/src/client/authProxyConfigId.integration.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      expect(outcome.ok).toBe(false)
+      expect((outcome as { error: Error }).error.message).toContain(
+        String(status),
+      )
+    })
+  }
+})
+
+describe('auth proxy config id: a 200 that carries no usable id', () => {
```
brtkx, line 126: I think we should also `it.todo` this one instead of asserting. I think we should add the following line to [authProxy.ts](https://github.com/zerodevapp/zerodev-wallet-sdk/blob/main/packages/core/src/client/authProxy.ts) as a fix. cc @SahilVasava 
```typescript
if (!authProxyConfigId) {
  throw new Error('Auth proxy config id is missing; cannot address the auth proxy.')
}
```

.github/workflows/ci.yml
christopherferreira9: Good catch! Updating it!

packages/core/src/adapters/walletAddressResolution.integration.test.ts
christopherferreira9: Can you explain a little further? I do consider these to be actual expected failures as it highlights broken code.

packages/core/src/storage/sessionTransition.integration.test.ts
brtkx: I think the only way we end up with read throwing is a bricked adapter from the dev. Our own storage should never throw given our implementation is sound; so propagating should be fine.

packages/core/src/core/registerMidFlowFailure.integration.test.ts
brtkx: The expectation is for user to retry manually. The only thing we don't have is separate error messaging based on which part has failed - either register passkey or login. But I think that would involve some rewrite to how errors are handled, which may be not worth it.

packages/core/src/adapters/walletAddressResolution.integration.test.ts
christopherferreira9: Settled offline, resolving this comment.

packages/core/src/core/loginWalletIdentity.integration.test.ts
brtkx: I don't think it should. For registration we need the guard to avoid duplicate wallets getting registered. For login, a duplicate call would only cause a new session being created. I don't think we should guard login e.g. if a bricked session needs to be replaced. We don't gain much by guarding it, but can introduce some edge cases.

packages/core/src/adapters/walletAddressResolution.integration.test.ts
brtkx: This is a great point because KMS is actually working on a feature to allow multiple wallets. This is something we will have to adjust in SDK; although not today.

packages/core/src/client/kmsFailureFidelity.integration.test.ts
brtkx: imo we don't need to back off and retry. trying once should be enough

packages/core/src/adapters/authorizationPayload.integration.test.ts
brtkx: https://linear.app/offchain-labs/issue/FS-2735/check-if-auth-address-is-a-valid-address-when-signing-authorization

Created a ticket. I wouldn't fix it as part of this PR so you don't have to publish a package.

packages/core/src/core/logoutRevocation.integration.test.ts
brtkx: Created a ticket.
https://linear.app/offchain-labs/issue/FS-2742/make-logout-errors-propagate-to-the-caller

packages/core/src/core/oauthSessionAbsent.integration.test.ts
brtkx: As of today, all KMS responses carry a session. This could change as part of custom OAuth sign-in, but I will clarify this one. As of now nothing is actionable.

packages/core/src/core/sessionReconciliation.integration.test.ts
brtkx: Yes, clearing every session should be expected for those cases

brtkx (approved):

packages/core/src/actions/wallet/userOperationPayload.integration.test.ts
brtkx: I already created a ticket as part of the previous question regarding validating the address. We have a few other validations in the sdk, but for the most part kms is going to do the final full validation.

packages/core/src/actions/wallet/userOperationPayload.integration.test.ts
brtkx: afaik KMS should check chain ID against chains allowed by the dev

packages/core/src/adapters/authorizationPayload.integration.test.ts
brtkx: I don't think there's much benefit for us to check those in the sdk. The KMS is already doing this check for us.

What I think we could do is stop taking `typedDataHash` as an argument and hash it ourselves in the sdk. However, this is a breaking change.

Created an issue: https://linear.app/offchain-labs/issue/ZER-1061/breaking-change-drop-typeddatahash-from-signtypeddatav4ts

PR #426, opened 2026-08-25
ci: only bump peer dependents when leaving the range

brtkx:
### Why packages got bumped to 1.0.0
By default, if peer dependency gets a minor/major bump, the consumer package gets a major bump. Because `wallet-react` has `wallet-core: workspace:*` as its peer dependency, and `wallet-core` got bumped to `0.1.0`, that is treated as a breaking change on the dev's side by the changeset.

### Workaround
https://github.com/changesets/changesets/blob/main/docs/experimental-options.md#onlyupdatepeerdependentswhenoutofrange-type-boolean
Setting this to true will only bump peer dependencies' major version if we go outside the range. I've set the range to `workspace:>=0.0.3` so the build passes. After `0.1.0` release we can bump it to `workspace:>=0.1.0`. Once we release `1.0.0` we can just get rid of this config.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -11,5 +11,8 @@
   "privatePackages": {
     "version": false,
     "tag": false
+  },
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
   }
 }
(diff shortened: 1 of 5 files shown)


brtkx:
Superseded by https://github.com/zerodevapp/zerodev-wallet-sdk/pull/428

PR #429, opened 2026-08-26
fix(wallet-ui): stop applying a global CSS reset to the host app

brtkx:
Using `@import "tailwindcss" prefix(zd)` in our wallet includes tailwind preflight, which resets the host app's global styling. Now we import what we need from `tailwindcss` and apply manual resets in `reset.css` for `zd-scope`. This way the host app is not affected at all.

### How to reproduce
Apply some global styles e.g. to `html` or `h1` tags in the demo app. Then import wallet-ui styles:
```typescript
import '@zerodev/wallet-react-ui/styles.css'
```

The global styles should be wiped now.


OmarBasem (approved):

OmarBasem:
@brtkx I found an issue in the UI as shown in the screenshot (might need to clean cache and rebuild to see it), pushed a fix in [this commit](https://github.com/zerodevapp/zerodev-wallet-sdk/pull/429/changes/f2a0f75c22c94fec111fc458424b46ed5e7736ac#diff-332f43c33ddf21e051af30f82cd4e869c92bfcea1915b5b7883b58e50d66e07e)

Screenshot 2026-08-27 at 9 04 45 AM

PR #431, opened 2026-08-27
fix(sra): stop applying a global CSS reset to the host app

OmarBasem:
Follow-up to https://github.com/zerodevapp/zerodev-wallet-sdk/pull/429


packages/react-ui/src/reset.css
brtkx, line 1: Don't we need to add `zd-scope` to the SRA component to apply the reset?

OmarBasem: @brtkx No, nothing needs adding. The SRA widget is already inside the scope.                                                                                                                                          
                                                                                                                                                                                                                 
`SmartRoutingAddress`, the package's only exported UI component, renders `<Screen>` as its root at `pages/index.tsx:127`, and `Screen` puts `zd-scope` on its outer div at `Screen/index.tsx:65`. Everything the widget renders is a descendant of that, so the reset applies with no per-component class.

brtkx (approved):

PR #440, opened 2026-08-31
fix(wallet-react-ui): verify link race condition

OmarBasem:
Closes [FS-2593](https://linear.app/offchain-labs/issue/FS-2593/verify-link-race-condition)

 ### Summary

  Opening a magic link (`/verify?code=…`) stripped the code from the URL and left a blank screen, even with a valid pending sign-in.

  **Root cause:** `Verifying` checks the store for the OTP session once on mount, and strips the code if it's missing. But the session restore (`auth.initialize()`, a sync localStorage read) only ran after `await connector.setup?.()`, whose async work (wallet creation, network session validation) is slower than the first render, so valid links were treated as dead.

  ### Changes
  
  - **Connector**: restore the OTP session *before* base setup, it now runs synchronously inside wagmi's `createConfig`, before any React effect can observe the store.
  - **`Verifying`**: with genuinely no session (link reused/expired), show a "Link Expired" error screen instead of a blank screen.
  - **Demo apps**: "Choose another sign-in method" on `/verify` now resets auth and redirects to `/`, where the home login reopens the widget.

### DIFF
--- a/.changeset/verify-otp-restore-race.md
+++ b/.changeset/verify-otp-restore-race.md
@@ -1,0 +1,19 @@
+---
+"@zerodev/wallet-react-ui": patch
+---
+
+fix: magic-link verification no longer dead-ends on `/verify?code=…`
+
+- The persisted OTP session is now restored before base connector setup.
+  Verification raced the restore: `auth.initialize()` (a synchronous
+  localStorage read) only ran after `await connector.setup?.()`, whose async
+  work (wallet creation, session validation) could be slow or fail.
+  `Verifying` reads the session once on mount, so landing on the magic link
+  before the restore finished stripped the code from the URL and silently
+  dropped the flow. The restore now runs first, synchronously inside wagmi's
+  `createConfig`, so the session is always in the store before any React
+  effect can observe it.
+- Opening a magic link with no pending OTP session (expired, already used,
+  or a browser that never started the email flow) now shows a "Link
+  Expired" error with a path back to sign-in, instead of stripping the code
+  and rendering nothing.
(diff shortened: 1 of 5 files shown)


brtkx:
[Codex] M1 | Medium | Add regression tests for restoration timing and missing-session recovery

Please add tests that fail when either fix is reverted. The current targeted tests accept both the fixed and broken implementations.

- [connector.test.ts:260](https://github.com/zerodevapp/zerodev-wallet-sdk/blob/8d53f3fe5e6dcd7033a845df82a9c0e04918dc37/packages/wallet-react-ui/src/connector.test.ts#L260) checks OTP restoration only after `await connector.setup()`. It passes whether restoration happens before or after base setup.
- [auth/index.test.tsx:38](https://github.com/zerodevapp/zerodev-wallet-sdk/blob/8d53f3fe5e6dcd7033a845df82a9c0e04918dc37/packages/wallet-react-ui/src/auth/index.test.tsx#L38) mocks `Verifying`, so it does not exercise the new missing-session screen or its retry button.

I checked commit `8d53f3f` in an isolated copy. All 52 existing targeted tests and three temporary checks passed. Restoring `connector.ts` and `Verifying.tsx` to their pre-PR versions kept those 52 tests green. Two temporary checks then failed on late OTP restoration and the absent "Link Expired" screen.

This is a coverage gap. A later refactor could move restoration below the await and reject a valid link while setup is slow, without failing these tests. I found no production defect in the changed code. Live email/KMS and browser navigation were not tested.

1. Hold base setup pending and assert both OTP fields immediately after calling setup. Suggested test shape:

   ```ts
   baseSetup.mockReturnValue(pendingSetup)
   const setup = connector.setup()

   expect(store.getState().auth.otpId).toBe('otp-stored')
   expect(store.getState().auth.otpEncryptionTargetBundle)
     .toBe('bundle-stored')

   resolveSetup()
   await setup
   ```

2. Render the real `Verifying` component with a code but no session. Assert that it shows "Link Expired", skips verification, and that clicking "Choose another sign-in method" removes `code` and selects `sign-up`.

brtkx (approved):

PR #445, opened 2026-09-02
fix: json-canonicalize broken publish + wagmi v2 dead wallet clicks

OmarBasem:
Two consumer-blocking fixes found while integrating the kit into Arbitrum Portal (external-wallets-only, wagmi 2 host).

## 1. `fix(core)`: pin `json-canonicalize` to exactly 2.0.0

`json-canonicalize@2.0.1` is a broken npm publish — its `main`/`module` point at `bundles/index.umd.js` / `esm5/index.js`, but the tarball only ships `src/*.ts` (verify: `curl -s $(npm view json-canonicalize@2.0.1 dist.tarball) | tar -tz`). Our `^2.0.0` range resolves to 2.0.1 on **every fresh install**, so new consumers of `@zerodev/wallet-core` crash at runtime with `Cannot find module 'json-canonicalize'`, three packages deep with no hint the SDK is involved. Our own lockfile holds 2.0.0, which is why it never reproduced here.

Repro:
```bash
npm i json-canonicalize@2.0.1 && node -e "require('json-canonicalize')"   # MODULE_NOT_FOUND
npm i json-canonicalize@2.0.0 && node -e "require('json-canonicalize')"   # ok
```

Pinned to exact 2.0.0. Upstream issue to be filed at snowyu/json-canonicalize.ts; 3.0.0 requires separate validation before adopting.

## 2. `fix(wallet-ui)`: wallet clicks were a silent no-op on wagmi v2 (commit f5d752b9)

Five components destructured `const { mutate: connect } = useConnect()` — valid only on wagmi v3, which spreads the raw TanStack mutation into the return. wagmi v2 strips `mutate`/`mutateAsync` and exposes only the `connect`/`connectAsync` aliases, so on v2 hosts every click on an installed wallet threw an uncaught `TypeError: ... is not a function` before `eth_requestAccounts` was sent (and with WC configured, the pairing preload could throw on `SignUp` mount). No install-time signal: the `^2.19.0 || ^3.0.0` peer range is *satisfied* by v2, and the SDK typechecks against its own installed v3.

Fix: destructure `connect` directly — it's wagmi's own alias for the same mutation function on **both** majors. Verified end-to-end on wagmi 2.19.5 (Arbitrum Portal): click → `wallet_requestPermissions` → `eth_requestAccounts` → connected.

Files: `SignUp/Wallet.tsx`, `SignUp/InstalledWallets.tsx`, `SignUp/MoreWallets.tsx`, `WalletSheet/index.tsx`, `useWalletConnectPairing.ts`.

**Follow-up suggestion (not in this PR):** add a wagmi 2.19 host to CI with one connect-flow test, so the dual-major peer promise is enforced mechanically.

Changesets included for both packages (patch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/pin-json-canonicalize.md
+++ b/.changeset/pin-json-canonicalize.md
@@ -1,0 +1,15 @@
+---
+'@zerodev/wallet-core': patch
+---
+
+fix: pin `json-canonicalize` to exactly 2.0.0
+
+`json-canonicalize@2.0.1` is a broken npm publish: its `main`/`module` point
+at `bundles/` and `esm5/` build output that is missing from the tarball (only
+`src/*.ts` ships). Our previous `^2.0.0` range resolved to 2.0.1 on every
+fresh install, so new consumers of `@zerodev/wallet-core` crashed at runtime
+with `Cannot find module 'json-canonicalize'`. The SDK repo's own lockfile
+held 2.0.0, which is why this never reproduced locally.
+
+Pinned to the exact known-good 2.0.0 until upstream ships a fixed release
+(3.0.0 requires separate validation before adopting).
--- a/.changeset/wagmi-v2-useconnect.md
+++ b/.changeset/wagmi-v2-useconnect.md
@@ -1,0 +1,17 @@
+---
+'@zerodev/wallet-react-ui': patch
+---
+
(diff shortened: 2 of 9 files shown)


kimkim12345678 (commented):
Nice

PR #458, opened 2026-09-11
fix(wallet-react-ui): clear step history when the auth flow ends

OmarBasem:
Closes [FS-2691
](https://linear.app/offchain-labs/issue/FS-2691/clear-step-history-when-the-auth-flow-ends)

### Summary                                                                                                                                                                                                        

Signing in with an external wallet and logging out left a back arrow on the sign-up screen that led nowhere.

### Demo

#### Before

https://github.com/user-attachments/assets/9780cca9-73ab-49d4-bccd-544f3ed6fd56



#### After
                                                                                          
       

https://github.com/user-attachments/assets/3fb6c387-5fad-4c05-81d4-5e0efb1e9d28

<!-- av pr metadata
This information is embedded by the av CLI when creating PRs to track the status of stacks when using Aviator. Please do not delete or edit this section of the PR.
```
{"parent":"main","parentHead":"","trunk":"main"}
```
-->


apps/zerodev-signer-demo/src/app/components/playground/utils/snippet.ts
```diff
@@ -91,6 +91,12 @@ export function emailMethod(
 export const DEFAULT_ITEMS: PlaygroundItem[] = [
   { key: 'default-passkey', type: 'passkey' },
   { key: 'default-divider', type: 'divider' },
+  {
```
brtkx: We shouldn't add this. The demo doesn't include 3P wallets by default.

brtkx (approved):

PR #461, opened 2026-09-14
fix(wallet-react-ui): handle wallet holding an earlier request

OmarBasem:
Closes [FS-2694](https://linear.app/offchain-labs/issue/FS-2694/say-when-the-wallet-still-holds-an-earlier-connection-request)

### Summary

This PR fixes the following scenario:

- Choose to login with MetaMask > Refresh with page without accepting > Choose MetaMask again
-> current behaviour: gives an error
-> expected behaviour: should display there is currently a pending request in MetaMask, then should connect upon accepting the request

#### Before

https://github.com/user-attachments/assets/74506e36-3d46-441f-8b04-dcb0792ffcf1

#### After

https://github.com/user-attachments/assets/257a3180-3c89-4c41-a8fa-edcd13f52ea7


packages/wallet-react-ui/src/auth/utils/isRequestPendingError.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * EIP-1193 `-32002`: the wallet already has a request open for this origin
```
brtkx: https://eips.ethereum.org/EIPS/eip-1193
I don't see any mention of `-32002` in this EIP. 

This error code is described by [EIP-1474](https://eips.ethereum.org/EIPS/eip-1474#error-codes); which only should return `Requested resource not found`. 

The message we are getting must be because the previous wallet connection cannot be found, but we can't assume there's an actual pending connection.

What we should do is parse this error properly (so we don't show `[object Object]`) and display it as is. Reown does the same thing - we should always just relay the error from the provider, without making our own assumptions in between.

What we can do however, is add our own note. We should display the error as is, and we can add something like:
`Open ${walletName} and check for an outstanding request. If one is waiting, approve or dismiss it, then try again.`

This way we don't claim that it's definitely the problem.

Let's also fix the comment so it doesn't mention EIP-1193.

brtkx (changes requested):

packages/wallet-react-ui/src/auth/utils/isRequestPendingError.ts
OmarBasem: @brtkx good points, fixed!

packages/wallet-react-ui/src/auth/pages/WalletConnecting.tsx
```diff
@@ -32,11 +33,48 @@ function describeConnectError(
...
+  if (isRequestPendingError(err)) {
+    // Usually nothing failed: the wallet still has the first request open
+    // and will not prompt again until it is answered there. But -32002 is
+    // only "resource unavailable", so suggest the outstanding request rather
+    // than assert it. Not the wallet's own message: MetaMask's names RPC
+    // methods and the origin ("Request of type 'wallet_requestPermissions'
+    // already pending for origin…") — Reown shows its own copy here too.
+    return {
+      title: `Check ${walletName}`,
+      message:
```
brtkx, line 48: I'd still show the original error message alongside our custom message - just before it. Probably need to reuse the object parsing block below - maybe we can put this parsing into utils?

Also probably would rename to `isMaybeRequestPendingError` because we shouldn't put guarantee on that in terms of semantics imo.

brtkx (approved):

PR #465, opened 2026-09-16
feat(core): add support for server wallets

brtkx:
## Summary

- Adds `@zerodev/wallet-core/server`: a client for a server process holding an agent key, covering wallet creation, listing, and the four sign operations on a KMS wallet set.
- Owner and admin operations (registering a wallet set, managing agent keys) are not part of the SDK; they need a dashboard session or the owner key.
- Web client behavior and public types are unchanged.

## Changes

**Server entry (`src/server/`)**
- `createServerWalletClient({ organizationId, privateKey | stamper })`: `privateKey` builds a stamper internally; `stamper` (any `SigningStamper`) wins when both are given.
- Exports `createPrivateKeyStamper`, `generateP256KeyPair`, and the action types. Nothing server-related is added to the root entry.

**Actions (`src/actions/serverWallet/`)**
- `createServerWallet`, `listServerWallets`, `listWalletSets`, `signMessage`, `signTransaction`, `signTypedDataV4`, `signUserOperation`, bound through `client/decorators/serverWallet.ts`.
- Sign parameters are the web types minus `token`; list responses mirror the KMS `WalletSetDetails` struct.
- Every call stamps under `X-Agent-Stamp`, the header the KMS reads agent stamps from; the stamp relayed to Turnkey inside the body keeps Turnkey's `X-Stamp` name.
- Lists stamp the `X-Timestamp` value themselves rather than through the transport, so the header name is the agent one.

**Stampers and client types**
- `SigningStamper` (`getPublicKey`, `sign`) split out of `ApiKeyStamper`, which keeps its rotation methods.
- `Client`, `ClientConfig`, `createBaseClient` gain a stamper type parameter defaulting to `ApiKeyStamper`, so existing `Client` references are unchanged; the server client is `Client<ServerWalletActions, SigningStamper>`.
- `Transport` accepts `SigningStamper`, the only capability it uses.
- `createPrivateKeyStamper(privateKey)`: P-256 API-key stamps from a raw hex key. `generateP256KeyPair()` returns compressed public key and private scalar as hex.

**Shared stamp encoding (`utils/utils.ts`)**
- `encodeStamp` (Turnkey `{ publicKey, scheme, signature }` envelope, base64url) and `compactSignatureToDerHex` move out of `indexedDbStamper` so both API-key stampers share one definition; `base64UrlEncode` accepts `Uint8Array` as well as `ArrayBuffer`.
- Scheme string and stamp header names live in `constants.ts`.

**Signing reuse**
- `sendSigningRequest`: `token` optional, bearer header only when present; outer header is `X-Agent-Stamp` for `server-wallet/` paths, otherwise the stamp's own name.

**Housekeeping**
- `package.json` exports `./server`; `biome.json` allows the two new barrel files; README gains a Server Wallets section.

## Breaking changes

- None. `Transport` implementations receive `SigningStamper` instead of `ApiKeyStamper`; existing ones compile since the web stamper satisfies both. No public name changes.


packages/core/README.md
```diff
@@ -326,6 +326,37 @@ await exportIframeStamper.injectKeyExportBundle(exportBundle, organizationId, 'H
...
+// At runtime
+const client = createServerWalletClient({
+  organizationId: WALLET_SET_SUB_ORG_ID,
+  privateKey: process.env.AGENT_PRIVATE_KEY,
+})
+
+const { walletAddress } = await client.createServerWallet({ projectId })
+const signature = await client.signMessage({
+  projectId,
+  organizationId: WALLET_SET_SUB_ORG_ID,
```
brtkx, line 353: @SahilVasava This is the same behavior with the web wallet, and since I'm reusing the types I need to keep `organizationId` required too. But I think we should have a follow up PR to make `organizationId` optional for both client and server wallets - `organizationId` is already part of `client` which we could reuse. I just didn't want to add more diff for now.

packages/core/src/client/types.ts
```diff
@@ -17,30 +21,38 @@ export type TransportConfig = {
 }
 
 export type Transport = (options: {
-  apiKeyStamper: ApiKeyStamper
+  apiKeyStamper: SigningStamper
```
brtkx, line 24: `ApiKeyStamper` contains all the key rotation params. `Transport` should not need it.

packages/core/src/stampers/types.ts
```diff
@@ -30,9 +30,26 @@ export type IframeStamper = Stamper & {
   applySettings(settings: { styles?: Record<string, string> }): Promise<boolean>
 }
 
-export type ApiKeyStamper = Stamper & {
```
brtkx, line 33: Stamper types are split into two stamper types now:
1. `SigningStamper` - for signing only.
2. `ApiKeyStamper` - more extended signer, used when key rotation is required. This is not needed for `Transport` or server wallets; that's the entire reason for the split.

packages/core/src/client/createClient.ts
brtkx, line 1: Suggesting having a look at `stampers/types.ts` before this one for more context.

packages/core/src/client/types.ts
brtkx, line 1: Suggesting having a look at `stampers/types.ts` before this one for more context.

packages/core/src/client/types.ts
```diff
@@ -17,30 +21,38 @@ export type TransportConfig = {
...
   passkeyStamper: PasskeyStamper
 }) => {
   config: TransportConfig
   request: RestRequestFn
   value?: Record<string, unknown>
 }
 
-export type ClientConfig = {
+export type ClientConfig<TStamper extends SigningStamper = ApiKeyStamper> = {
```
brtkx, line 32: By default it's `ApiKeyStamper` with key rotation etc, so that we don't have to change the existing stamper logic at all. Server wallets override this, because they don't need key rotation.

packages/core/src/stampers/indexedDbStamper.ts
brtkx, line 1: Reusing some utils between server wallets. Check changes to `utils/utils.ts`. 99% of the changes are just moving stuff to shared utils.

packages/core/src/server/createServerWalletClient.ts
```diff
@@ -0,0 +1,60 @@
...
+ * transport as the web client; only the stamper in the slot and the bound
+ * actions differ. Owner and admin operations are not here: they belong to the
+ * dashboard.
+ */
+export function createServerWalletClient(
+  config: ServerWalletClientConfig,
+): ServerWalletClient {
+  const stamper = resolveStamper(config)
+  return createBaseClient<undefined, SigningStamper>({
+    transport: zeroDevWalletTransport({ baseUrl: `${KMS_SERVER_URL}/api/v1` }),
```
SahilVasava: This pins the client to prod. Server-wallet routes are behind `feat:server-wallets`, which is only in `staging.values.yaml`, so as shipped every call 404s. The web client takes `proxyBaseUrl`; a `url` option here (plus `timeoutMs` / `fetchOptions` passthrough) is needed before anyone can run this against staging.

packages/core/src/actions/serverWallet/createServerWallet.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+  const timestampMs = Date.now()
+
+  const innerStamp = await client.apiKeyStamper.stamp(
+    canonicalizeEx(createWalletPayload(organizationId, timestampMs)),
+  )
+  const body = {
+    timestampMs,
+    stamp: {
+      stampHeaderName: innerStamp.stampHeaderName,
```
SahilVasava: The embedded header name comes from the stamper. KMS validates it as `oneof=X-Stamp X-Stamp-Webauthn` and uses it to pick the Turnkey header, so it has to be `X-Stamp` regardless of what `stamp()` returns. `createPrivateKeyStamper` happens to return `X-Stamp`, but the config accepts any `SigningStamper` (the HSM case) and nothing on that type says so; a stamper returning `X-Agent-Stamp` gets a 400. Suggest `stampHeaderName: TURNKEY_STAMP_HEADER` here and in `sendSigningRequest`.

Longer term the SDK could build the stamp itself from `getPublicKey()` + `sign()` so custom stampers never construct envelopes. Today a custom stamper can't build a correct one anyway: `encodeStamp` and `compactSignatureToDerHex` aren't exported from `./server`.

packages/core/src/actions/wallet/signingUtils.ts
```diff
@@ -82,14 +92,17 @@ export async function sendSigningRequest(
   // Outer stamp over full body (for KMS middleware)
   const fullBodyString = canonicalizeEx(fullBody)
   const outerStamp = await client.apiKeyStamper.stamp(fullBodyString)
+  const outerHeader = path.startsWith(SERVER_WALLET_PREFIX)
```
SahilVasava: Same point as in `createServerWallet`: line 87 embeds `innerStamp.stampHeaderName`, which must be `X-Stamp` for KMS to accept it. Hardcode `TURNKEY_STAMP_HEADER` for the inner name.

Nit: the `server-wallet/` prefix check couples this shared helper to a URL shape. An explicit `outerStampHeader` param (default `outerStamp.stampHeaderName`) that the server actions pass as `AGENT_STAMP_HEADER` says the same thing without the string match.

packages/core/src/actions/serverWallet/createServerWallet.ts
```diff
@@ -0,0 +1,76 @@
...
+  walletId: string
+  walletAddress: Hex
+}
+
+/**
+ * Turnkey CREATE_WALLET activity, byte-for-byte what the KMS rebuilds from the
+ * request (`ext/turnkey/create_user.go` CreateWalletPayload). The KMS relays the
+ * agent's stamp over it and co-signs with its own key.
+ */
+function createWalletPayload(organizationId: string, timestampMs: number) {
```
SahilVasava: This is the byte-sensitive part: Turnkey verifies the agent's stamp over whatever KMS reconstructs, so any drift here fails with an opaque `failed to create wallet on turnkey`. I generated the golden from Go's `CreateWalletPayload(1700000000000, "org-123")` and it matches the current output, but nothing pins it (this file is at 0% in codecov). Same for `createPrivateKeyStamper`: no test that its stamp actually verifies. Both pass with the following, feel free to drop it in:

```ts
import { createPublicKey, verify as nodeVerify } from 'node:crypto'
import { p256 } from '@noble/curves/nist.js'
import { canonicalizeEx } from 'json-canonicalize'
import { describe, expect, it, vi } from 'vitest'
import { createServerWallet } from '../actions/serverWallet/createServerWallet.js'
import type { Client } from '../client/types.js'
import { createPrivateKeyStamper } from '../stampers/privateKeyStamper.js'
import type { SigningStamper } from '../stampers/types.js'
import { generateP256KeyPair } from '../utils/p256KeyPair.js'

// json.Marshal(CreateWalletPayload(1700000000000, "org-123")) from doorway-kms ext/turnkey/create_user.go
const GO_CREATE_WALLET_PAYLOAD = JSON.parse(
  `{"organizationId":"org-123","parameters":{"accounts":[{"addressFormat":"ADDRESS_FORMAT_ETHEREUM","curve":"CURVE_SECP256K1","path":"m/44'/60'/0'/0/0","pathFormat":"PATH_FORMAT_BIP32"}],"walletName":"default wallet-1700000000000"},"timestampMs":"1700000000000","type":"ACTIVITY_TYPE_CREATE_WALLET"}`,
)

function decodeStamp(value: string) {
  return JSON.parse(Buffer.from(value, 'base64url').toString('utf8')) as {
    publicKey: string
    scheme: string
    signature: string
  }
}

describe('createPrivateKeyStamper', () => {
  it('produces a Turnkey stamp whose signature is ECDSA-P256/SHA-256 DER over the payload', async () => {
    const { privateKey, publicKey } = generateP256KeyPair()
    const stamp = await createPrivateKeyStamper(privateKey).stamp('{"hello":"world"}')
    expect(stamp.stampHeaderName).toBe('X-Stamp')

    const envelope = decodeStamp(stamp.stampHeaderValue)
    expect(envelope.scheme).toBe('SIGNATURE_SCHEME_TK_API_P256')
    expect(envelope.publicKey).toBe(publicKey)

    const raw = p256.Point.fromHex(envelope.publicKey).toBytes(false)
    const key = createPublicKey({
      format: 'jwk',
      key: {
        kty: 'EC',
        crv: 'P-256',
        x: Buffer.from(raw.slice(1, 33)).toString('base64url'),
        y: Buffer.from(raw.slice(33, 65)).toString('base64url'),
      },
    })
    expect(
      nodeVerify('sha256', Buffer.from('{"hello":"world"}'), { key, dsaEncoding: 'der' }, Buffer.from(envelope.signature, 'hex')),
    ).toBe(true)
  })
})

describe('createServerWallet', () => {
  it('stamps the same CREATE_WALLET activity the KMS rebuilds', async () => {
    vi.spyOn(Date, 'now').mockReturnValue(1700000000000)
    const stamped: string[] = []
    const request = vi.fn(async () => ({ walletId: 'w', walletAddress: '0x1' }))
    const client = {
      organizationId: 'org-123',
      apiKeyStamper: {
        stamp: vi.fn(async (payload: string) => {
          stamped.push(payload)
          return { stampHeaderName: 'X-Stamp', stampHeaderValue: 'stamp' }
        }),
      },
      request,
    } as unknown as Client<undefined, SigningStamper>

    await createServerWallet(client, { projectId: 'project' })

    expect(stamped[0]).toBe(canonicalizeEx(GO_CREATE_WALLET_PAYLOAD))
    const [call] = request.mock.calls[0] as unknown as [{ headers: Record<string, string>; body: { stamp: { stampHeaderName: string } } }]
    expect(stamped[1]).toBe(canonicalizeEx(call.body))
    expect(call.headers['X-Agent-Stamp']).toBe('stamp')
    expect(call.body.stamp.stampHeaderName).toBe('X-Stamp')
  })
})
```

packages/core/README.md
```diff
@@ -326,6 +326,37 @@ await exportIframeStamper.injectKeyExportBundle(exportBundle, organizationId, 'H
...
+} from '@zerodev/wallet-core/server'
+
+// Once, when provisioning: the public key is registered on the wallet set by
+// its owner, the private key goes into the server's secrets.
+const { publicKey, privateKey } = generateP256KeyPair()
+
+// At runtime
+const client = createServerWalletClient({
+  organizationId: WALLET_SET_SUB_ORG_ID,
+  privateKey: process.env.AGENT_PRIVATE_KEY,
```
SahilVasava: `process.env.AGENT_PRIVATE_KEY` is `string | undefined`, so this example fails under `strict`. Needs a `!` or a guard.

SahilVasava (changes requested):
Verified rather than just read: core typecheck and the 303 core tests pass on the branch, `./server` builds and imports in Node (ESM and CJS), a stamp from `createPrivateKeyStamper` verifies against KMS's own `VerifyAPIKeyStamp` in Go, and the `createServerWallet` inner payload matches Go's `CreateWalletPayload` byte-for-byte. Stamping the canonical body and sending the non-canonical one is fine since KMS canonicalizes before hashing.

Two things to fix before merge, both inline: the client can't be pointed at staging, where the routes actually exist, and the inner stamp header name is taken from the stamper instead of fixed to `X-Stamp`. Plus a test suggestion for the two byte-sensitive paths and two nits.

packages/core/src/actions/serverWallet/createServerWallet.ts
brtkx: Similar tests are part of my [tests commit](https://github.com/zerodevapp/zerodev-wallet-sdk/pull/465/commits/0b2d8c47bdb7b05ef7afb242f1516e281a2e4199) I've just pushed.

packages/core/README.md
brtkx: IMO readme doesn't need to be an exact working snippet. We could have put something like `'YOUR_AGENT_PRIVATE_KEY'` string literal in the example, which wouldn't work either out of the box.

packages/core/src/actions/serverWallet/createServerWallet.ts
brtkx: Addressed https://github.com/zerodevapp/zerodev-wallet-sdk/pull/465/changes/14a9de6fa231e788bb84b16fd972b41fea9e51b1

packages/core/src/server/createServerWalletClient.ts
brtkx: Addressed https://github.com/zerodevapp/zerodev-wallet-sdk/pull/465/commits/c17dbacf8d7c3de9708e1003fa0a52d91e161be3

packages/core/src/actions/wallet/signingUtils.ts
brtkx: Addressed https://github.com/zerodevapp/zerodev-wallet-sdk/pull/465/changes/14a9de6fa231e788bb84b16fd972b41fea9e51b1 https://github.com/zerodevapp/zerodev-wallet-sdk/pull/465/commits/747718e58b88fa5be740b55d654e4f5c98455b0a

SahilVasava (approved):
All addressed. Checked the new head: `proxyBaseUrl` and `fetchOptions` reach the transport, the inner stamp name is fixed to `X-Stamp` in both `sendSigningRequest` and `createServerWallet`, the prefix check is gone in favour of `outerStampHeader`, and the new tests pin the stamper to the RFC 6979 vector and the CREATE_WALLET fixture to what the KMS rebuilds (matches the Go output I generated). The README note about ACL is right too: the KMS matches Origin, Referer or client IP and rejects a request with none of them allowlisted, so a server has to send an allowlisted Origin or be IP-allowlisted.

